PR #1, opened 2016-07-13
Copy/paste fix

moderndeveloperllc:
Looks to be a simple copy/paste error. On a side note, can't wait for this to be up!


moderndeveloperllc:
fixed a few days ago it looks like

PR #3, opened 2016-10-27
Remove migrations-php.yaml in favor of copying entire directory

jmikola:
For local builds, this removes the need to manually list each source file in the migrations YAML file, which is bound to have problems once we multiple branches.


Makefile
```diff
@@ -18,6 +18,12 @@ help:
 	@echo 'Variables'
 	@echo '  ARGS         - Arguments to pass to mut-publish'
 
+html:
+	rm -r source/
+	mkdir source
+	cp -a build/php-library/docs/* source
+	giza make html
```
i80and: giza make publish

Makefile
```diff
@@ -18,6 +18,10 @@ help:
 	@echo 'Variables'
 	@echo '  ARGS         - Arguments to pass to mut-publish'
 
+publish:
+	rsync -a --delete build/php-library/docs/ source/
```
i80and: Ehhhhh shoot I should have expected that. How do you folks feel about using git submodules instead of giza's assets system?

For example, you would run `git submodule add -f -b master -- https://github.com/mongodb/mongo-php-library.git build/php-library`, and then `git submodule update` right before the `rsync`

schmalliso:
Merged!

PR #5, opened 2017-08-25
DRIVERS-409: Update copyright blurb

jmikola:
https://jira.mongodb.org/browse/DRIVERS-409

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -39,7 +39,7 @@ source_suffix = '.txt'
 master_doc = sconf.master_doc
 language = 'en'
 project = sconf.project
-copyright = u'2008-{0}'.format(datetime.date.today().year)
+copyright = u'2008-present'
 
 version = conf.version.branch
 release = conf.version.release


conf.py
```diff
@@ -39,7 +39,7 @@
 master_doc = sconf.master_doc
 language = 'en'
 project = sconf.project
-copyright = u'2008-{0}'.format(datetime.date.today().year)
+copyright = u'MongoDB, Inc. 2008-present. MongoDB, Mongo, and the leaf logo are registered trademarks of MongoDB, Inc'
```
kay-kim: You just need the string to be 2008-present. 
We add all the MongoDB blurb in the jinja template when we build the html automatically.

kay-kim: Just a little background -- because python docs don't use the jinja template, their conf.py is a little different, whereas since php-library uses our jinja template, just the date range is all we need.

jmikola: Fixed!

kay-kim: Works for me!  👍

PR #16, opened 2020-07-31
DOCSP-11542: Update version in build_conf to correct title

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-11548

Staging:
https://docs-mongodbcom-staging.corp.mongodb.com/php-library/ccho/DOCSP-11542-update-v1.7-conf/reference/class/MongoDBCollection.html#phpclass.MongoDB\Collection

From some experimentation, it looks like config/build_conf.yaml contains an entry for version that needed to be updated for your v1.7 branch.

### DIFF
--- a/config/build_conf.yaml
+++ b/config/build_conf.yaml
@@ -10,7 +10,7 @@ project:
   title: "PHP Library Manual"
   branched: true
 version:
-  release: '1.6'
+  release: '1.7'
   branch: 'master'
 system:
   runstate:


config/build_conf.yaml
```diff
@@ -10,7 +10,7 @@ project:
   title: "PHP Library Manual"
   branched: true
 version:
-  release: '1.6'
+  release: '1.7'
```
jmikola, line 13: Is this the default version?

I'm curious if this means that the `master` (upcoming) branch will also be reported as 1.7. Now that we have a v1.7 branch in PHPLIB, the `master` branch is technically `1.8-dev`.

ccho-mongodb: I think it will need an update for master branch, will look at what other repos are doing.

ccho-mongodb: Confirmed with the docs platform team that this is a label, so I'll create a separate PR for master to show 1.8-dev

PR #57, opened 2023-12-15
Update manual and php extlinks

jmikola:
Use https for both links and update "manual" to use the canonical docs URL

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -56,8 +56,8 @@ rst_epilog = '\n'.join([
 
 extlinks = {
     'issue': ('https://jira.mongodb.org/browse/%s', '' ),
-    'manual': ('http://docs.mongodb.org/manual%s', ''),
-    'php': ('http://php.net/%s', '')
+    'manual': ('https://www.mongodb.com/docs/manual%s', ''),
+    'php': ('https://php.net/%s', '')
 }
 
 ## add `extlinks` for each published version.
--- a/conf.py
+++ b/conf.py
@@ -57,6 +57,7 @@ rst_epilog = '\n'.join([
 extlinks = {
     'issue': ('https://jira.mongodb.org/browse/%s', '' ),
     'manual': ('https://www.mongodb.com/docs/manual%s', ''),
+    'atlas': ('https://www.mongodb.com/docs/atlas%s', ''),
     'php': ('https://php.net/%s', '')
 }
 


GromNaN (approved):

GromNaN:
I suppose it would have no effect if you merged it on older branches, as the documentation is published only for tags?

ccho-mongodb:
> I suppose it would have no effect if you merged it on older branches, as the documentation is published only for tags?

I think you can backport this to all the other non-legacy version branches on this repo. The older version branches (other than for legacy versions which are read-only on the docs platform) should be publishable.

conf.py
```diff
@@ -56,8 +56,8 @@
 
 extlinks = {
     'issue': ('https://jira.mongodb.org/browse/%s', '' ),
-    'manual': ('http://docs.mongodb.org/manual%s', ''),
-    'php': ('http://php.net/%s', '')
+    'manual': ('https://www.mongodb.com/docs/manual%s', ''),
```
GromNaN, line 59: Could you add the `docs` prefix for the need of https://github.com/mongodb/mongo-php-library/pull/1273#discussion_r1559471310

```suggestion
    'manual': ('https://www.mongodb.com/docs/manual%s', ''),
    'docs': ('https://www.mongodb.com/docs%s', ''),
```

jmikola: Would `:atlas:` make more sense? I see that is used in some existing docs outside of PHPLIB. For instance, the "Connect from Your Application" link in [Python Drivers: Compatibility](https://www.mongodb.com/docs/drivers/python-drivers/#compatibility) uses that (see: [here](https://github.com/mongodb/docs-ecosystem/blob/2a75928245e315b64f86f252d8cfcf6ad039e7f2/source/python-drivers.txt#L51)).

There are definitely other, existing links to https://www.mongodb.com/docs/atlas/ throughout the PHPLIB docs that could be changed to take advantage of that.

Will defer to @ccho-mongodb here.

GromNaN: Yes! I didn't think to check how it was configured elsewhere.
```suggestion
    'manual': ('https://www.mongodb.com/docs/manual%s', ''),
    'atlas': ('https://www.mongodb.com/docs/atlas%s', ''),
```

jmikola:
> I think you can backport this to all the other non-legacy version branches on this repo. The older version branches (other than for legacy versions which are read-only on the docs platform) should be publishable.

How would backports be handled? Cherry-pick after merging?

If the docs team is willing to handle that as owners of this repo I'm all for it. IIRC, Snooty is used for 1.16+ so it'd be three branches in addition to master.

ccho-mongodb:
> > I think you can backport this to all the other non-legacy version branches on this repo. The older version branches (other than for legacy versions which are read-only on the docs platform) should be publishable.
> 
> How would backports be handled? Cherry-pick after merging?
> 
> If the docs team is willing to handle that as owners of this repo I'm all for it. IIRC, Snooty is used for 1.16+ so it'd be three branches in addition to master.

Sure, happy to backport and deploy this change for you.

jmikola:
@ccho-mongodb: @schmalliso confirmed that `conf.py` does nothing for Snooty. I'll close this PR, but I think the right call is to remove this file from the v1.16 and newer branches.

PR #58, opened 2024-01-23
DOCSP-35780: enable snooty build

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-35780

Compare with the following files on the master branch:
- [snooty.toml](https://github.com/mongodb/docs-php-library/blob/master/snooty.toml)
- [worker.sh](https://github.com/mongodb/docs-php-library/blob/master/worker.sh)

### DIFF
--- a/snooty.toml
+++ b/snooty.toml
@@ -1,0 +1,19 @@
+name = "php-library"
+title = "PHP Library Manual"
+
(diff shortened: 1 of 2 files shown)


rustagir (approved):
lgtm

PR #106, opened 2024-08-15
DOCSP-41957 - Mongo Client

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-41957
Staging - https://preview-mongodbmongokart.gatsbyjs.io/php-library/docsp-41957-mongo-client/connect/client/

## Self-Review Checklist

- [ ] Are all the links working?


snooty.toml
```diff
@@ -24,3 +24,4 @@ php-library = "MongoDB PHP Library"
 
 [constants]
 php-library = "MongoDB PHP Library"
+driver-short = "PHP library"
```
mongoKart, line 27: using "driver" so existing source constant references keep working. can rename all refs in a later PR if needed

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+
+     - Required. A prefix that identifies this as a string in the
+       standard connection format.
+
+   * - ``username:password``
+
+     - Optional. Authentication credentials. If you include these, the client
+       authenticates the user against the database specified in ``authSource``.
+       For more information about the ``authSource`` connection option, see
+       :ref:`php-auth`.
```
mongoKart, line 60: broken for now

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+     - Optional. The authentication database to use if the
+       connection string includes ``username:password@``
+       authentication credentials but not the ``authSource`` option. If you don't include 
+       this component, the client authenticates the user against the ``admin`` database.
+
+   * - ``?<options>``
+
+     - Optional. A query string that specifies connection-specific
+       options as ``<name>=<value>`` pairs. See
+       :ref:`php-connection-options` for a full description of
```
mongoKart, line 78: broken for now

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+   database, most applications require only a single instance of
+   ``MongoDB\Client``, even across multiple requests.
+
+API Documentation
+-----------------
+
+To learn more about creating a ``MongoDB\Client`` object in the {+driver-short+},
+see the following API documentation:
+
+- :ref:`MongoDB\Client <php-api-mongodbclient>`
```
mongoKart, line 110: broken for now, possibly because this is staging

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+To connect to a MongoDB deployment, you need two things:
```
rustagir, line 23:
```suggestion
To connect to a MongoDB deployment, you must create the following items:
```

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+
+- A **connection URI**, also known as a *connection string*, which tells the {+driver-short+}
+  which MongoDB deployment to connect to.
+- A **MongoDB\\Client** object, which creates the connection to the MongoDB deployment
+  and lets you perform operations on it.
```
rustagir, line 28:
```suggestion
- **Connection URI**, also known as a *connection string*, which tells the {+driver-short+}
  which MongoDB deployment to connect to.
- **MongoDB\\Client** object, which creates the connection to the MongoDB deployment
  and lets you perform operations on it.
```

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+
+You can also use either of these components to customize the way the {+driver-short+} behaves
+while connected to MongoDB.
+
+This guide shows you how to create a connection string and use a ``MongoDB\Client`` object
+to connect to MongoDB.
```
rustagir, line 34: Q: does this guide show how to create a connection string or does it just teach about connection strings?

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+
+   * - Component
+     - Description
+
+   * - ``mongodb://``
+
+     - Required. A prefix that identifies this as a string in the
+       standard connection format.
+
+   * - ``username:password``
```
rustagir, line 55: S: change to ``db_username`` and ``db_password``

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+   * - ``?<options>``
+
+     - Optional. A query string that specifies connection-specific
+       options as ``<name>=<value>`` pairs. See
+       :ref:`php-connection-options` for a full description of
+       these options.
+
+For more information about creating a connection string, see
+:manual:`Connection Strings </reference/connection-string>` in the
+MongoDB Server documentation.
```
rustagir, line 83:
```suggestion
To learn more about connection strings, see
:manual:`Connection Strings </reference/connection-string>` in the
MongoDB Server documentation.
```

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+
+Create a ``MongoDB\Client``
+---------------------------
+
+To create a connection to MongoDB, construct an instance of the ``MongoDB\Client`` class,
+passing the connection URI as a string to the constructor.
```
rustagir, line 89:
```suggestion
To create a connection to MongoDB, pass your connection string when constructing
 an instance of the ``MongoDB\Client`` class.
```

source/connect/client.txt
rustagir, line 110: I: this is presented as an API link, but I see a ref, is this supposed to link here https://www.mongodb.com/docs/php-library/current/reference/class/MongoDBClient/?

rustagir (changes requested):

source/connect/client.txt
mongoKart: it describes the components of the connection string

source/connect/client.txt
mongoKart: do these need to be disambiguated from other credentials? the following paragraph also says they're for the DB.

source/connect/client.txt
mongoKart: yes, the PHP docs host their own API docs

source/connect/client.txt
rustagir: I think it could be either. Just figured we might start integrating those placeholders where applicable so that there are fewer mis-matches

rustagir: Have you considered including that connection string image on this page and using those placeholders instead? another option

source/connect/client.txt
rustagir: Hmm, inter-repo refs should work. Maybe you can double check the snooty-toml to make sure its included in the intersphinx?

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+
+Overview
+--------
+
+To connect to a MongoDB deployment, you must create the following items:
+
+- **Connection URI**, also known as a *connection string*, which tells the {+driver-short+}
+  which MongoDB deployment to connect to.
+- **MongoDB\\Client** object, which creates the connection to the MongoDB deployment
+  and lets you perform operations on it.
```
rustagir, line 28: Q: are there two \\ because it's being escaped? It looks normal on the frontend.

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+
+You can also use either of these components to customize the way the {+driver-short+} behaves
+while connected to MongoDB.
```
rustagir, line 31: S: maybe you can add some more information about what this suggests. Maybe:

```suggestion
You can also set options within either or both of these components to
customize the way that the {+driver-short+} behaves
while connected to MongoDB.
```

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+   * - ``?<options>``
+
+     - Optional. A query string that specifies connection-specific
+       options as ``<name>=<value>`` pairs. See
+       :ref:`php-connection-options` for a full description of
+       these options.
+
+To learn more about connection strings, see
+:manual:`Connection Strings </reference/connection-string>` in the
+MongoDB Server documentation.
```
rustagir, line 83:
```suggestion
Server manual.
```

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+
+Create a ``MongoDB\Client``
+---------------------------
+
+To create a connection to MongoDB, pass your connection string when constructing
+ an instance of the ``MongoDB\Client`` class.
```
rustagir, line 89: I: something about the indentation is causing this to render weirdly:
```suggestion
To create a connection to MongoDB, pass your connection string when constructing
an instance of the ``MongoDB\Client`` class.
```

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+     - Optional. A query string that specifies connection-specific
+       options as ``<name>=<value>`` pairs. See
+       :ref:`php-connection-options` for a full description of
+       these options.
+
+To learn more about connection strings, see
+:manual:`Connection Strings </reference/connection-string>` in the
+MongoDB Server documentation.
+
+Create a ``MongoDB\Client``
```
rustagir, line 85: S: avoid using monospace in titles: https://www.mongodb.com/docs/meta/style-guide/style/titles-and-headings/#:~:text=Don%27t%20apply%20font%20treatments%20(bold%2C%20italics%2C%20or%20monospace)%20to%20text%20in%20a%20title%20or%20heading.

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+
+.. literalinclude:: /includes/connect/client.php
+   :language: php
+   :copyable: true
+
+.. tip:: Reuse Your Client
+
+   Because each ``MongoDB\Client`` object represents a pool of connections to the
+   database, most applications require only a single instance of
+   ``MongoDB\Client``, even across multiple requests.
```
rustagir, line 102:
```suggestion
   ``MongoDB\Client`` that you can use across multiple requests.
```

rustagir (changes requested):
a few more small things!

source/connect/client.txt
mongoKart: yeah, you need to do this when it's bold for some reason

source/connect/client.txt
mongoKart: Do you think we need to be clearer about this point? I hoped that if we showed users all of the pieces of the connection string, they could put one together...?

source/connect/client.txt
mongoKart: I'll follow the Style Guide, but this is a Bad Rule imo

source/connect/client.txt
mongoKart: Not opposed to this change, but should we be consistent with the Server manual?

https://www.mongodb.com/docs/manual/reference/connection-string/#connection-string-components

source/connect/client.txt
mongoKart: Good thought, but this is within the same repo. Looks like my addition of the ref at the top of that file wasn't recognized, even though I had committed the change (I removed the last line of the file just so VS Code would let me save/commit again).

source/connect/client.txt
rustagir: I dont think many users are creating a connection string from its constituent pieces. Perhaps this sentence could express something more like:
```suggestion
This guide describes the components of a connection string and shows how to
 use a ``MongoDB\Client`` object to connect to a MongoDB deployment.
```

source/connect/client.txt
rustagir: Looks to be working now!

rustagir (approved):
lgtm

source/connect/client.txt
```diff
@@ -0,0 +1,110 @@
...
+deployment on port ``27017`` of ``localhost``:
+
+.. literalinclude:: /includes/connect/client.php
+   :language: php
+   :copyable: true
+
+.. tip:: Reuse Your Client
+
+   Because each ``MongoDB\Client`` object represents a pool of connections to the
+   database, you can use a single instance of the class across multiple requests.
```
GromNaN, line 102: To be checked by @jmikola @alcaeus, but I don't think there's a pool of connections, since requests are blocking and PHP is monothread.

source/includes/connect/client.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+$client = new MongoDB\Client("mongodb://localhost:27017");
```
GromNaN, line 3: Missing newline at end of file.
```suggestion
$client = new MongoDB\Client("mongodb://localhost:27017");

```

GromNaN (changes requested):

source/connect/client.txt
alcaeus: That is correct, PHP does not implement a connection pool. We have an [article in the extension docs](https://www.php.net/manual/en/mongodb.connection-handling.php) that details the connection persistence mechanism used by the PHP driver (current and the legacy one used before 1.2.0).

GromNaN (approved):
LGTM

source/connect/client.txt
jmikola: I'm a bit late here, but I don't think these should be `db_username` and `db_password`. DB-backed auth is but one mechanism. These could be credentials for a different auth mechanism, so more generic names are appropriate.

See also: [connection string spec](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.md)

PR #113, opened 2024-08-22
DOCSP-41981: Change streams

norareidy:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-41981
Staging - https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-41981-change-streams/read/change-streams/
## Self-Review Checklist

### DIFF
--- a/source/includes/read/change-streams.php
+++ b/source/includes/read/change-streams.php
@@ -1,10 +1,12 @@
 <?php
 require 'vendor/autoload.php'; 
 
+// start-to-json
 function toJSON(object $document): string
 {
     return MongoDB\BSON\Document::fromPHP($document)->toRelaxedExtendedJSON();
 }
+// end-to-json
 
 $uri = getenv('MONGODB_URI') ?: throw new RuntimeException('Set the MONGODB_URI variable to your Atlas URI that connects to the sample dataset');
 $client = new MongoDB\Client($uri);
--- a/source/includes/read/change-streams.php
+++ b/source/includes/read/change-streams.php
@@ -25,6 +25,10 @@ for ($changeStream->rewind(); true; $changeStream->next()) {
     }
     $event = $changeStream->current();
     echo toJSON($event) . PHP_EOL;
+
+    if ($event['operationType'] === 'invalidate') {
+        break;
+    }
 }
 // end-open-change-stream
 
@@ -47,6 +51,10 @@ for ($changeStream->rewind(); true; $changeStream->next()) {
     }
     $event = $changeStream->current();
     echo toJSON($event) . PHP_EOL;
+
+    if ($event['operationType'] === 'invalidate') {
+        break;
+    }
 }
 // end-change-stream-pipeline
 
@@ -61,6 +69,10 @@ for ($changeStream->rewind(); true; $changeStream->next()) {
     }
     $event = $changeStream->current();
     echo toJSON($event) . PHP_EOL;
+
+    if ($event['operationType'] === 'invalidate') {
+        break;
+    }
 }
 // end-change-stream-post-image
 
--- a/source/includes/read/change-streams.php
+++ b/source/includes/read/change-streams.php
@@ -13,7 +13,7 @@ $collection = $client->sample_restaurants->restaurants;
 $changeStream = $collection->watch();
 
 foreach ($changeStream as $event) {
-    echo json_encode($event), PHP_EOL;
+    echo json_encode($event) . PHP_EOL;
 }
 // end-open-change-stream
 
@@ -31,17 +31,17 @@ $pipeline = [['$match' => ['operationType' => 'update']]];
 $changeStream = $collection->watch($pipeline);
 
 foreach ($changeStream as $event) {
-    echo json_encode($event), PHP_EOL;
+    echo json_encode($event) . PHP_EOL;
 }
 // end-change-stream-pipeline
 
 // Passes an options argument to watch() to include the post-image of updated documents
 // start-change-stream-post-image
-$options = ['fullDocument' => 'updateLookup'];
+$options = ['fullDocument' => MongoDB\Operation\Watch::FULL_DOCUMENT_UPDATE_LOOKUP];
 $changeStream = $collection->watch([], $options);
 
 foreach ($changeStream as $event) {
-    echo json_encode($event), PHP_EOL;
+    echo json_encode($event) . PHP_EOL;
 }
 // end-change-stream-post-image
 
--- a/source/includes/read/change-streams.php
+++ b/source/includes/read/change-streams.php
@@ -1,0 +1,47 @@
+<?php
+require 'vendor/autoload.php'; 
+
+$uri = getenv('MONGODB_URI') ?: throw new RuntimeException('Set the MONGODB_URI variable to your Atlas URI that connects to the sample dataset');
+$client = new MongoDB\Client($uri);
+
+// start-db-coll
+$collection = $client->sample_restaurants->restaurants;
+// end-db-coll
+
+// Monitors and prints changes to the "restaurants" collection
+// start-open-change-stream
+$changeStream = $collection->watch();
+
+foreach ($changeStream as $event) {
+    echo json_encode($event), PHP_EOL;
+}
+// end-open-change-stream
+
+// Updates a document that has a "name" value of "Blarney Castle"
+// start-update-for-change-stream
+$result = $collection->updateOne(
+    ['name' => 'Blarney Castle'],
+    ['$set' => ['cuisine' => 'Irish']]
+);
+// end-update-for-change-stream
+
+// Passes a pipeline argument to watch() to monitor only update operations
+// start-change-stream-pipeline
+$pipeline = [['$match' => ['operationType' => 'update']]];
+$changeStream = $collection->watch($pipeline);
+
+foreach ($changeStream as $event) {
+    echo json_encode($event), PHP_EOL;
+}
+// end-change-stream-pipeline
+
+// Passes an options argument to watch() to include the post-image of updated documents
+// start-change-stream-post-image
+$options = ['fullDocument' => 'updateLookup'];
+$changeStream = $collection->watch([], $options);
+
+foreach ($changeStream as $event) {
+    echo json_encode($event), PHP_EOL;
+}
+// end-change-stream-post-image
+
--- a/source/includes/read/change-streams.php
+++ b/source/includes/read/change-streams.php
@@ -1,6 +1,11 @@
 <?php
 require 'vendor/autoload.php'; 
 
+function toJSON(object $document): string
+{
+    return MongoDB\BSON\Document::fromPHP($document)->toRelaxedExtendedJSON();
+}
+
 $uri = getenv('MONGODB_URI') ?: throw new RuntimeException('Set the MONGODB_URI variable to your Atlas URI that connects to the sample dataset');
 $client = new MongoDB\Client($uri);
 
@@ -12,8 +17,12 @@ $collection = $client->sample_restaurants->restaurants;
 // start-open-change-stream
 $changeStream = $collection->watch();
 
-foreach ($changeStream as $event) {
-    echo json_encode($event) . PHP_EOL;
+for ($changeStream->rewind(); true; $changeStream->next()) {
+    if ( ! $changeStream->valid()) {
+        continue;
+    }
+    $event = $changeStream->current();
+    echo toJSON($event) . PHP_EOL;
 }
 // end-open-change-stream
 
@@ -30,8 +39,12 @@ $result = $collection->updateOne(
 $pipeline = [['$match' => ['operationType' => 'update']]];
 $changeStream = $collection->watch($pipeline);
 
-foreach ($changeStream as $event) {
-    echo json_encode($event) . PHP_EOL;
+for ($changeStream->rewind(); true; $changeStream->next()) {
+    if ( ! $changeStream->valid()) {
+        continue;
+    }
+    $event = $changeStream->current();
+    echo toJSON($event) . PHP_EOL;
 }
 // end-change-stream-pipeline
 
@@ -40,8 +53,12 @@ foreach ($changeStream as $event) {
 $options = ['fullDocument' => MongoDB\Operation\Watch::FULL_DOCUMENT_UPDATE_LOOKUP];
 $changeStream = $collection->watch([], $options);
 
-foreach ($changeStream as $event) {
-    echo json_encode($event) . PHP_EOL;
+for ($changeStream->rewind(); true; $changeStream->next()) {
+    if ( ! $changeStream->valid()) {
+        continue;
+    }
+    $event = $changeStream->current();
+    echo toJSON($event) . PHP_EOL;
 }
 // end-change-stream-post-image
 
--- a/source/read/change-streams.txt
+++ b/source/read/change-streams.txt
@@ -1,0 +1,257 @@
+.. _php-change-streams:
+
+====================
+Monitor Data Changes
+====================
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. facet::
+   :name: genre
+   :values: reference
+
+.. meta::
+   :keywords: watch, code example
+
+Overview
+--------
+
+In this guide, you can learn how to use a **change stream** to monitor real-time
+changes to your database. A change stream is a {+mdb-server+} feature that
+allows your application to subscribe to data changes on a collection, database,
+or deployment.
+
+When using the {+driver-short+}, you can instantiate a ``MongoDB\ChangeStream`` to
+monitor data changes.
+
+Sample Data
+~~~~~~~~~~~
+
+The examples in this guide use the ``restaurants`` collection in the ``sample_restaurants``
+database from the :atlas:`Atlas sample datasets </sample-data>`. To access this collection
+from your PHP application, instantiate a ``MongoDB\Client`` that connects to an Atlas cluster
+and assign the following value to your ``collection`` variable:
+
+.. literalinclude:: /includes/read/change-streams.php
+   :language: php
+   :dedent:
+   :start-after: start-db-coll
+   :end-before: end-db-coll
+
+To learn how to create a free MongoDB Atlas cluster and load the sample datasets, see the
+:atlas:`Get Started with Atlas </getting-started>` guide.
+
+Open a Change Stream
+--------------------
+
+To open a change stream, call the ``watch()`` method. The instance on which you
+call the ``watch()`` method on determines the scope of events that the change
+stream listens for. You can call the ``watch()`` method on the following
+classes:
+
+- ``MongoDB\Client``: Monitor all changes in the MongoDB deployment
+- ``MongoDB\Database``: Monitor changes in all collections in the database
+- ``MongoDB\Collection``: Monitor changes in the collection
+
+The following example opens a change stream on the ``restaurants`` collection
+and outputs changes as they occur:
+
+.. literalinclude:: /includes/read/change-streams.php
+   :start-after: start-open-change-stream
+   :end-before: end-open-change-stream
+   :language: php
+   :dedent:
+
+To begin watching for changes, run the preceding code. Then, in a separate
+application or shell, modify the ``restaurants`` collection. The following
+example updates a document that has a ``name`` field value of ``'Blarney Castle'``:
+
+.. _php-change-stream-update:
+
+.. literalinclude:: /includes/read/change-streams.php
+   :start-after: start-update-for-change-stream
+   :end-before: end-update-for-change-stream
+   :language: php
+   :dedent:
+
+When you update the collection, the change stream application prints the change
+as it occurs. The printed change event resembles the following output:
+
+.. code-block:: bash
+   :copyable: false
+
+   { "_id" : { "_data" : "..." }, "operationType" : "update", "clusterTime" : 
+   { "$timestamp" : { ... }, "wallTime" : { "$date" : ... }, "ns" : 
+   { "db" : "sample_restaurants", "coll" : "restaurants" }, "documentKey" : 
+   { "_id" : { "$oid" : "..." } }, "updateDescription" : { "updatedFields" :
+   { "cuisine" : "Irish" }, "removedFields" : [  ], "truncatedArrays" : [  ] } }
+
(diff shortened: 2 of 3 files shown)


source/read/change-streams.txt
```diff
@@ -0,0 +1,254 @@
...
+Overview
+--------
+
+In this guide, you can learn how to use a **change stream** to monitor real-time
+changes to your data. A change stream is a {+mdb-server+} feature that
+allows your application to subscribe to data changes on a collection, database,
+or deployment.
+
+When using the {+php-library+}, you can instantiate a ``MongoDB\ChangeStream`` to
+monitor data changes.
```
rustagir, line 29:
```suggestion
When using the {+php-library+}, you can create an instance of ``MongoDB\ChangeStream`` to
monitor data changes, such as updates, insertions, and deletions.
```

source/read/change-streams.txt
```diff
@@ -0,0 +1,254 @@
...
+.. literalinclude:: /includes/read/change-streams.php
+   :start-after: start-update-for-change-stream
+   :end-before: end-update-for-change-stream
+   :language: php
+   :dedent:
+
+When you update the collection, the change stream application prints the change
+as it occurs. The printed change event resembles the following output:
+
+.. code-block:: bash
```
rustagir, line 84: S: change the type to `none` instead of bash, as this is output not a command. I think `console` might work too

source/read/change-streams.txt
```diff
@@ -0,0 +1,254 @@
...
+- ``$addFields`` or ``$set``: Adds new fields to documents
+- ``$match``: Filters the documents
+- ``$project``: Projects a subset of the document fields
+- ``$replaceWith`` or ``$replaceRoot``: Replaces the input document with the
+  specified document
+- ``$redact``: Restricts the contents of the documents
+- ``$unset``: Removes fields from documents
+
+The following passes a pipeline that includes the ``$match`` stage to the ``watch()``
+method. This instructs the ``watch()`` method to output only update operations:
```
rustagir, line 108:
```suggestion
The following passes a pipeline that includes the ``$match`` stage to the ``watch()``
method. This instructs the ``watch()`` method to output events only when update operations occur:
```

source/read/change-streams.txt
```diff
@@ -0,0 +1,254 @@
...
+   :start-after: start-change-stream-pipeline
+   :end-before: end-change-stream-pipeline
+   :language: php
+   :dedent:
+
+Modify ``watch()`` Behavior
+---------------------------
+
+To modify the behavior of the ``watch()`` method, you can pass an options array
+as a parameter to ``watch()``. The following table describes some options you
```
rustagir, line 120:
```suggestion
as a parameter to ``watch()``. The following table describes useful options you
```

source/read/change-streams.txt
```diff
@@ -0,0 +1,254 @@
...
+   :widths: 30 70
+   :header-rows: 1
+
+   * - Option
+     - Description
+
+   * - ``fullDocument``
+     - | Specifies whether to show the full document after the change, rather
+         than showing only the changes made to the document. To learn more about
+         this option, see :ref:`php-change-stream-pre-post-image`.
```
rustagir, line 133:
```suggestion
         this option, see the :ref:`php-change-stream-pre-post-image` section of this guide.
```

source/read/change-streams.txt
```diff
@@ -0,0 +1,254 @@
...
+    
+   * - ``resumeAfter``
+     - | Instructs ``watch()`` to resume returning changes after the
+         operation specified in the resume token.
+       | Each change stream event document includes a resume token as the ``_id``
+         field. Pass the entire ``_id`` field of the change event document that
+         represents the operation you want to resume after.
+       | This option is mutually exclusive with ``startAfter`` and ``startAtOperationTime``.
```
rustagir, line 146: S: since you mention `startAtOperationTime`, maybe you can include it in this table

source/read/change-streams.txt
```diff
@@ -0,0 +1,254 @@
...
+.. literalinclude:: /includes/read/change-streams.php
+   :start-after: start-change-stream-post-image
+   :end-before: end-change-stream-post-image
+   :language: php
+   :dedent:
+
+With the change stream application running, updating a document in the
+``restaurants`` collection by using the :ref:`preceding update example
+<php-change-stream-update>` prints a change event resembling the following
+code:
```
rustagir, line 219:
```suggestion
output:
```

rustagir (changes requested):
a few small things!

rustagir (approved):
lgtm!

source/includes/read/change-streams.php
```diff
@@ -0,0 +1,47 @@
+<?php
+require 'vendor/autoload.php'; 
+
+$uri = getenv('MONGODB_URI') ?: throw new RuntimeException('Set the MONGODB_URI variable to your Atlas URI that connects to the sample dataset');
```
jmikola, line 4: In all of the original PHPLIB code examples, we fall back to `mongodb://127.0.0.1/` (see: [aggregate.php](https://github.com/mongodb/mongo-php-library/blob/1.19.1/examples/aggregate.php#L23)). Why not do so here?

source/includes/read/change-streams.php
```diff
@@ -0,0 +1,47 @@
...
+$client = new MongoDB\Client($uri);
+
+// start-db-coll
+$collection = $client->sample_restaurants->restaurants;
+// end-db-coll
+
+// Monitors and prints changes to the "restaurants" collection
```
jmikola, line 11: Are these comments intended for the readers, or just the doc authors. Asking as it appears outside of the ``start`` and ``end`` blocks used for including code in the rendered docs.

source/includes/read/change-streams.php
```diff
@@ -0,0 +1,47 @@
...
+// start-open-change-stream
+$changeStream = $collection->watch();
+
+foreach ($changeStream as $event) {
```
jmikola, line 15: `foreach` loops are not suitable for iterating tailable cursors (such as those returned for change streams), as the loop will terminate as soon as there are no documents to consume. In other words, `foreach` prevents us from actually polling the cursor and waiting for additional data.

This is why the PHPLIB docs had a [Tailable Cursor Iteration](https://www.mongodb.com/docs/php-library/current/tutorial/tailable-cursor/) tutorial. The existing examples in the [`watch()` API reference](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBCollection-watch/#examples) demonstrate an equivalent approach, with added code to break out of the loop should we receive an "invalidate" event.

I'm not sure if this script is ever intended to be executed in its entirety, or simply serve as an assortment of snippets to include in the rendered docs. But if you're using the proper approach to iterate the change stream, note that the first loop is unlikely to terminate on its own and that would prevent subsequent statements in the file from executing.

source/includes/read/change-streams.php
```diff
@@ -0,0 +1,47 @@
...
+    echo json_encode($event) . PHP_EOL;
```
jmikola, line 16: Since the events are BSON documents, it'd be preferable to use the existing BSON functions to output JSON. Relying on PHP's `json_encode()` functions will not produce Extended JSON for types that need a special representation.

This is why the existing examples in PHPLIB use a common help function:

```php
function toJSON(object $document): string
{
    return MongoDB\BSON\Document::fromPHP($document)->toRelaxedExtendedJSON();
}
```

source/read/change-streams.txt
```diff
@@ -0,0 +1,261 @@
...
+Overview
+--------
+
+In this guide, you can learn how to use a **change stream** to monitor real-time
+changes to your data. A change stream is a {+mdb-server+} feature that
+allows your application to subscribe to data changes on a collection, database,
+or deployment.
+
+When using the {+php-library+}, you can create an instance of ``MongoDB\ChangeStream`` to
+monitor data changes, such as updates, insertions, and deletions.
```
jmikola, line 29: Users never construct a `MongoDB\ChangeStream` object. It'd be more correct to say they can invoke the various `watch()` methods on Client, Database, or Collection objects to obtain a ChangeStream object that they can then iterate to monitor data changes on the logical entity (cluster, database, or collection, respectively).

source/read/change-streams.txt
```diff
@@ -0,0 +1,261 @@
...
+
+Sample Data
+~~~~~~~~~~~
+
+The examples in this guide use the ``restaurants`` collection in the ``sample_restaurants``
+database from the :atlas:`Atlas sample datasets </sample-data>`. To access this collection
+from your PHP application, instantiate a ``MongoDB\Client`` that connects to an Atlas cluster
+and assign the following value to your ``collection`` variable:
```
jmikola, line 37: `$client->sample_restaurants->restaurants` constructs a new collection, so it'd be more accurate to say something like "select a collection to assign to the ``collection`` variable" or "instantiate your ``collection`` variable".

Also, since you're referring to a variable by name here, would it make sense to use ``$collection``? That would be consistent with how use a "()" suffix when referring to methods like `watch()` below.

source/read/change-streams.txt
```diff
@@ -0,0 +1,261 @@
...
+When you update the collection, the change stream application prints the change
+as it occurs. The printed change event resembles the following output:
+
+.. code-block:: none
+   :copyable: false
+
+   {"_id":{"_data":"..."},"operationType":"update","clusterTime":{"$timestamp":{...},
+   "wallTime":{"$date":...},"ns":{"db":"sample_restaurants","coll":"restaurants"},
+   "documentKey":{"_id":{"$oid":"..."}},"updateDescription":{"updatedFields":
+   {"cuisine":"Irish"},"removedFields":[],"truncatedArrays":[]}}}
```
jmikola, line 90: If `json_encode()` is used in the PHP example, I don't think you'll see keys like "$timestamp", "$date", and "$oid" in the output here. The BSON API will need to be used to ensure Extended JSON syntax is produced.

source/read/change-streams.txt
```diff
@@ -0,0 +1,261 @@
...
+         than showing only the changes made to the document. To learn more about
+         this option, see the :ref:`php-change-stream-pre-post-image` section of this
+         guide.
+
+   * - ``fullDocumentBeforeChange``
+     - | Specifies whether to show the full document as it was before the change, rather
+         than showing only the changes made to the document. To learn more about
+         this option, see :ref:`php-change-stream-pre-post-image`.
+    
+   * - ``resumeAfter``
```
jmikola, line 142: Note: I assume this change stream tutorial is common to other driver docs, so consider this a more general question beyond the scope of PHPLIB.

Is `resumeAfter` actually relevant to the users? I know it's the original option and used internally by drivers when resuming, but my understanding is that `startAfter` is preferred for users.

Looking at [`db.collection.watch()`](https://www.mongodb.com/docs/manual/reference/method/db.collection.watch/), which documents both options, it looks like `resumeAfter` only differs by its limitation to _not_ allow resuming after an "invalidate" event. Given that, it doesn't seem like a "useful option" to include in this table (although it absolutely belongs in the API reference).

source/read/change-streams.txt
```diff
@@ -0,0 +1,261 @@
...
+         field. Pass the entire ``_id`` field of the change event document that
+         represents the operation you want to resume after.
+       | This option is mutually exclusive with ``resumeAfter`` and ``startAtOperationTime``.
+
+   * - ``startAtOperationTime``
+     - | Instructs the change stream to only provide changes that occurred at or after
+         the specified timestamp. 
+       | This option is mutually exclusive with ``startAfter`` and ``resumeAfter``.
+
+   * - ``typeMap``
```
jmikola, line 164: This also seems like something we can omit from this page if the table is admittedly a subset of "useful options". It isn't demonstrated in any examples, and it's already covered in the API reference.

source/read/change-streams.txt
```diff
@@ -0,0 +1,261 @@
...
+The **pre-image** is the full version of a document *before* a change. To include the
+pre-image in the change stream event, set the ``fullDocumentBeforeChange`` option
+to one of the following values:
+
+- ``MongoDB\Operation\Watch::FULL_DOCUMENT_BEFORE_CHANGE_WHEN_AVAILABLE``: The change event includes
+  a pre-image of the modified document for change events. If the pre-image is not available, this
+  change event field has a ``null`` value.
+- ``MongoDB\Operation\Watch::FULL_DOCUMENT_BEFORE_CHANGE_REQUIRED``: The change event includes a pre-image
+  of the modified document for change events. If the pre-image is not available, the
+  {+php-library+} raises an error.
```
jmikola, line 200: Technically, the server raises errors for this and the corresponding post-image case. PHPLIB would just propagate that error to the user. Not sure if this is important to clarify, as this might mislead users into thinking it's a client-side error.

source/read/change-streams.txt
```diff
@@ -0,0 +1,261 @@
...
+  {+php-library+} raises an error.  
+
+The **post-image** is the full version of a document *after* a change. To include the
+post-image in the change stream event, set the ``fullDocument`` option to
+one of the following values:
+
+- ``MongoDB\Operation\Watch::FULL_DOCUMENT_UPDATE_LOOKUP``: The change event includes a
+  copy of the entire changed document from some time after the change.
+- ``MongoDB\Operation\Watch::FULL_DOCUMENT_WHEN_AVAILABLE``: The change event includes
+  a post-image of the modified document for change events only if the post-image is available.
```
jmikola, line 209: Should this clarify that `null` is used as a filler value, as you do for pre-image?

source/includes/read/change-streams.php
norareidy: The message about setting the MONGODB_URI variable was a request from Jerome on a previous PR, but I'm fine with either option. If you have a strong preference for `mongodb://127.0.0.1/` I'm happy to change it

source/includes/read/change-streams.php
norareidy: These comments are for AI ingestion to clarify what the code's doing. We try not to make them visible on the page if possible

source/read/change-streams.txt
norareidy: Fixed!

source/includes/read/change-streams.php
jmikola: Happy to leave as-is if this @GromNaN already signed off on this approach.

source/read/change-streams.txt
norareidy: Changed to `$collection`! Do you mean that `$client->sample_restaurants->restaurants` creates a new `restaurants` collection rather than accessing the `restaurants` sample collection? From running the code, it does query the sample data

source/read/change-streams.txt
norareidy: If startAfter is preferred, I'm happy to remove resumeAfter

source/includes/read/change-streams.php
norareidy: I changed these loops to for loops and edited based on the examples you linked - let me know what you think

source/includes/read/change-streams.php
jmikola: Noted that you added the `toJSON()` utility function to the docs.

source/read/change-streams.txt
jmikola: My objection was with the word "value" in "assign the following value" because `$client->sample_restaurants->restaurants` is not actually a value even though it looks like we're accessing PHP properties.

The code here is using the [`Database::__get()`](https://github.com/mongodb/mongo-php-library/blob/1.19.1/src/Database.php#L163) method, which delegates to [`Database::selectCollection()`](https://github.com/mongodb/mongo-php-library/blob/1.19.1/src/Database.php#L559). That method constructs a new `MongoDB\Collection` object each time it is called. To me, "value" implies that we're assigning some fixed value.

If it's kosher to write "assign the following to your `$collection` variable:" I'd be in favor of that. But if "value" needs to stay we can resolve this thread without further changes.

Note: In the PHPLIB docs I've historically used the `Client::selectDatabase()`, `Client::selectCollection()`, and `Database::selectCollection()` methods; however, I understand that you might want to use the shorthand syntax for consistency with other driver/shell docs.

> Do you mean that $client->sample_restaurants->restaurants creates a new restaurants collection rather than accessing the restaurants sample collection?

Creation of the `MongoDB\Collection` object has no impact on the server. We're just initializing the object with strings to identify the database and collection (i.e. namespace). The driver will only create a collection on the server if we explicitly invoke the [`create` command](https://www.mongodb.com/docs/manual/reference/command/create/) or write to a collection that doesn't exist (implicit creation).

source/read/change-streams.txt
jmikola: Did you mean to change the text here to clarify that PHPLIB will propagate the server-side error?

source/includes/read/change-streams.php
GromNaN: This examples are using the [sample data](https://www.mongodb.com/docs/atlas/sample-data/) from Atlas. Do you know if their is any documented way to get this data locally?

source/includes/read/change-streams.php
```diff
@@ -0,0 +1,66 @@
...
+// Passes a pipeline argument to watch() to monitor only update operations
+// start-change-stream-pipeline
+$pipeline = [['$match' => ['operationType' => 'update']]];
+$changeStream = $collection->watch($pipeline);
+
+for ($changeStream->rewind(); true; $changeStream->next()) {
+    if ( ! $changeStream->valid()) {
+        continue;
+    }
+    $event = $changeStream->current();
```
jmikola, line 48: Note: add conditional break somewhere in this loop for the "invalidate" event.

source/includes/read/change-streams.php
```diff
@@ -0,0 +1,66 @@
...
+// Passes an options argument to watch() to include the post-image of updated documents
+// start-change-stream-post-image
+$options = ['fullDocument' => MongoDB\Operation\Watch::FULL_DOCUMENT_UPDATE_LOOKUP];
+$changeStream = $collection->watch([], $options);
+
+for ($changeStream->rewind(); true; $changeStream->next()) {
+    if ( ! $changeStream->valid()) {
+        continue;
+    }
+    $event = $changeStream->current();
```
jmikola, line 62: Note: add conditional break somewhere in this loop for the "invalidate" event.

source/read/change-streams.txt
```diff
@@ -0,0 +1,268 @@
...
+--------
+
+In this guide, you can learn how to use a **change stream** to monitor real-time
+changes to your data. A change stream is a {+mdb-server+} feature that
+allows your application to subscribe to data changes on a collection, database,
+or deployment.
+
+When using the {+php-library+}, you can call the ``watch()`` method to return an
+instance of ``MongoDB\ChangeStream``. Then, you can iterate through change events stored 
+in the ``MongoDB\ChangeStream`` instance to monitor data changes, such as updates,
```
jmikola, line 30: Unless you need to include the phrase "change events" here, consider "...iterate through the ``MongoDB\ChangeStream`` instance to..." as a more concise alternative.

I don't think of cursor objects as something that store data, so the original phrasing as odd to me. Ignore this feedback if this is the same language you're using in other language docs, though.

source/read/change-streams.txt
```diff
@@ -0,0 +1,268 @@
...
+
+.. _php-change-stream-update:
+
+.. literalinclude:: /includes/read/change-streams.php
+   :start-after: start-update-for-change-stream
+   :end-before: end-update-for-change-stream
+   :language: php
+   :dedent:
+
+When you update the collection, the change stream application prints the change
```
jmikola, line 95: Something that just occurred to me: should this tutorial advise users to run this PHP code through a shell/terminal? Users would likely run into issues trying to execute the `watch()` process through a web server.

I'm curious if there's language from other language tutorials (assuming this was adapted from existing content) we can use.

The most straightforward way to address this might be to consistently refer to "shell process" instead of "application or shell" for all of the code examples (no harm in using that for the update example as well).

source/read/change-streams.txt
```diff
@@ -0,0 +1,268 @@
...
+The following example calls the ``watch()`` method on a collection and includes the post-image
+of updated documents by setting the ``fullDocument`` option:
+
+.. literalinclude:: /includes/read/change-streams.php
+   :start-after: start-change-stream-post-image
+   :end-before: end-change-stream-post-image
+   :language: php
+   :dedent:
+
+With the change stream application running, updating a document in the
```
jmikola, line 227: I think "updating" should be "update".

Per my earlier comment about pointing users away from using a web server, the following might be clearer:

> With the change stream application running in a separate shell, update a document...

source/includes/read/change-streams.php
jmikola: Sorry, I confused these with the data sets we already discuss in the PHPLIB docs: https://www.mongodb.com/docs/php-library/current/tutorial/example-data/

I poked around https://www.mongodb.com/docs/atlas/sample-data/ and it didn't see any way to download those directly :disappointed:

source/read/change-streams.txt
norareidy: changed!

source/read/change-streams.txt
norareidy: oops yes I did, fixed now

source/read/change-streams.txt
norareidy: Added "in a separate shell" - "using" makes more grammatical sense in this case

source/read/change-streams.txt
norareidy: We use "application" pretty consistently across all docs, but I agree that it makes sense to suggest running the application from the shell. I updated all mentions of running the application on this page to clarify that

source/includes/read/change-streams.php
```diff
@@ -0,0 +1,66 @@
...
+// Monitors and prints changes to the "restaurants" collection
+// start-open-change-stream
+$changeStream = $collection->watch();
+
+for ($changeStream->rewind(); true; $changeStream->next()) {
+    if ( ! $changeStream->valid()) {
+        continue;
+    }
+    $event = $changeStream->current();
+    echo toJSON($event) . PHP_EOL;
```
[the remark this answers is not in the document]
norareidy: changed & tested

source/read/change-streams.txt
jmikola: > "updating" makes more grammatical sense in this case

My mistake. I missed that the sentence continued with "prints a change event..." after the `:ref:` link.

jmikola (approved):

PR #115, opened 2024-08-27
DOCSP-41988: Aggregation

norareidy:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-41988
Staging - https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-41988-aggregation/aggregation/

## Self-Review Checklist

### DIFF
--- a/source/aggregation.txt
+++ b/source/aggregation.txt
@@ -94,8 +94,8 @@ Aggregation Example
    free MongoDB Atlas cluster and load the sample datasets, see the :atlas:`Get Started with Atlas
    </getting-started>` guide.
 
-To perform an aggregation, pass an array containing the aggregation pipeline
-stages to the ``MongoDB\Collection::aggregate()`` method.
+To perform an aggregation, pass an array containing the pipeline stages to
+the ``MongoDB\Collection::aggregate()`` method.
 
 The following code example produces a count of the number of bakeries in each borough
 of New York. To do so, it uses an aggregation pipeline that contains the following stages:
@@ -136,13 +136,12 @@ An execution plan is a potential way in which MongoDB can complete an operation.
 When you instruct MongoDB to explain an operation, it returns both the
 plan MongoDB executed and any rejected execution plans.
 
-To explain an aggregation operation, run the ``explain`` database command by passing
-the command information to the ``MongoDB\Database::command()`` method. You must specify the
-``aggregate``, ``pipeline``, and ``cursor`` fields in the ``explain`` command document
-to explain the aggregation.
+To explain an aggregation operation, construct a ``MongoDB\Operation\Aggregate`` object
+and pass the database, collection, and pipeline stages as parameters. Then, pass the
+``MongoDB\Operation\Aggregate`` object to the ``MongoDB\Collection::explain()`` method.
 
-The following example instructs MongoDB to explain the aggregation operation from the
-preceding :ref:`php-aggregation-example`:
+The following example instructs MongoDB to explain the aggregation operation
+from the preceding :ref:`php-aggregation-example`:
 
 .. io-code-block::
    :copyable:
@@ -161,7 +160,6 @@ preceding :ref:`php-aggregation-example`:
       "planCacheKey":"D56D6F10","optimizedPipeline":true,"maxIndexedOrSolutionsReached":false,
       "maxIndexedAndSolutionsReached":false,"maxScansToExplodeReached":false,"winningPlan":{
       ... }
-
 
 Additional Information
 ----------------------
--- a/source/aggregation.txt
+++ b/source/aggregation.txt
@@ -108,6 +108,7 @@ of New York. To do so, it uses an aggregation pipeline that contains the followi
   value
 
 .. io-code-block::
+   :copyable:
 
    .. input:: /includes/aggregation.php
       :start-after: start-match-group
@@ -116,7 +117,8 @@ of New York. To do so, it uses an aggregation pipeline that contains the followi
       :dedent:
 
    .. output:: 
-
+      :visible: false
+      
       {"_id":"Brooklyn","count":173}
       {"_id":"Queens","count":204}
       {"_id":"Bronx","count":71}
@@ -143,6 +145,7 @@ The following example instructs MongoDB to explain the aggregation operation fro
 preceding :ref:`php-aggregation-example`:
 
 .. io-code-block::
+   :copyable:
 
    .. input:: /includes/aggregation.php
       :start-after: start-explain
@@ -151,6 +154,7 @@ preceding :ref:`php-aggregation-example`:
       :dedent:
 
    .. output:: 
+      :visible: false
 
       {"explainVersion":"2","queryPlanner":{"namespace":"sample_restaurants.restaurants",
       "indexFilterSet":false,"parsedQuery":{"cuisine":{"$eq":"Bakery"}},"queryHash":"865F14C3",
--- a/source/aggregation.txt
+++ b/source/aggregation.txt
@@ -166,6 +166,11 @@ preceding :ref:`php-aggregation-example`:
 Additional Information
 ----------------------
 
+To view a tutorial that uses the {+php-library+} to create complex aggregation
+pipelines, see `Complex Aggregation Pipelines with Vanilla PHP and MongoDB 
+<https://www.mongodb.com/developer/products/mongodb/aggregations-php-mongodb/>`__
+in the MongoDB Developer Center.
+
 MongoDB Server Manual
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/source/aggregation.txt
+++ b/source/aggregation.txt
@@ -67,7 +67,7 @@ You can use aggregation operations to perform the following actions:
 Limitations
 ~~~~~~~~~~~
 
-Keep the following limitations in mind when using aggregation operations:
+Consider the following limitations when performing aggregation operations:
 
 - Returned documents cannot violate the
   :manual:`BSON document size limit </reference/limits/#mongodb-limit-BSON-Document-Size>`
@@ -76,11 +76,11 @@ Keep the following limitations in mind when using aggregation operations:
   limit by creating an options array that sets the ``allowDiskUse`` option to ``true``
   and passing the array to the ``MongoDB\Collection::aggregate()`` method.
 
-.. important:: $graphLookup Exception
+  .. important:: $graphLookup Exception
 
-   The :manual:`$graphLookup
-   </reference/operator/aggregation/graphLookup/>` stage has a strict
-   memory limit of 100 megabytes and ignores the ``allowDiskUse`` option.
+     The :manual:`$graphLookup
+     </reference/operator/aggregation/graphLookup/>` stage has a strict
+     memory limit of 100 megabytes and ignores the ``allowDiskUse`` option.
 
 .. _php-aggregation-example:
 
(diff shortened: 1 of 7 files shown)


source/aggregation.txt
```diff
@@ -0,0 +1,197 @@
...
+- Run find operations
+- Rename fields
+- Calculate fields
+- Summarize data
+- Group values
+
+Limitations
+~~~~~~~~~~~
+
+Keep the following limitations in mind when using aggregation operations:
```
jordan-smith721, line 70: Nit:
```suggestion
Consider the following limitations when performing aggregation operations:
```

source/aggregation.txt
```diff
@@ -0,0 +1,197 @@
...
+  of 16 megabytes.
+- Pipeline stages have a memory limit of 100 megabytes by default. You can exceed this
+  limit by creating an options array that sets the ``allowDiskUse`` option to ``true``
+  and passing the array to the ``MongoDB\Collection::aggregate()`` method.
+
+.. important:: $graphLookup Exception
+
+   The :manual:`$graphLookup
+   </reference/operator/aggregation/graphLookup/>` stage has a strict
+   memory limit of 100 megabytes and ignores the ``allowDiskUse`` option.
```
jordan-smith721, line 83: I think this admonition should be indented since it applies to the second bullet above

source/aggregation.txt
```diff
@@ -0,0 +1,197 @@
...
+      {"_id":"Missing","count":2}
+      {"_id":"Manhattan","count":221}
+
+Explain an Aggregation
+~~~~~~~~~~~~~~~~~~~~~~
+
+To view information about how MongoDB executes your operation, you can
+instruct the MongoDB query planner to **explain** it. When MongoDB explains
+an operation, it returns **execution plans** and performance statistics.
+An execution plan is a potential way MongoDB can complete an operation.
```
jordan-smith721, line 135:
```suggestion
An execution plan is a potential way in which MongoDB can complete an operation.
```

source/aggregation.txt
```diff
@@ -0,0 +1,197 @@
...
+When you instruct MongoDB to explain an operation, it returns both the
+plan MongoDB executed and any rejected execution plans.
+
+To explain an aggregation operation, run the ``explain`` database command by passing
+the command information to the ``MongoDB\Database::command()`` method. You must set the
+``aggregate``, ``pipeline``, and ``cursor`` fields of the ``explain`` command document
+to explain the aggregation.
```
jordan-smith721, line 142: S: The word "set" here made me think you were going to explain what these fields needed to be set to
```suggestion
the command information to the ``MongoDB\Database::command()`` method. You must specify the
``aggregate``, ``pipeline``, and ``cursor`` fields in the ``explain`` command document
to explain the aggregation.
```

source/aggregation.txt
```diff
@@ -0,0 +1,197 @@
...
+Additional Information
+----------------------
+
+MongoDB Server Manual
+~~~~~~~~~~~~~~~~~~~~~
+
+To view a full list of expression operators, see :manual:`Aggregation
+Operators. </reference/operator/aggregation/>`
+
+To learn about assembling an aggregation pipeline and view examples, see
```
jordan-smith721, line 175:
```suggestion
To learn about assembling an aggregation pipeline and to view examples, see
```

source/aggregation.txt
```diff
@@ -0,0 +1,197 @@
...
+      "planCacheKey":"D56D6F10","optimizedPipeline":true,"maxIndexedOrSolutionsReached":false,
+      "maxIndexedAndSolutionsReached":false,"maxScansToExplodeReached":false,"winningPlan":{
+      ... }
+
+
+Additional Information
+----------------------
+
+MongoDB Server Manual
+~~~~~~~~~~~~~~~~~~~~~
```
jordan-smith721, line 170: I think this would work better as a short sentence followed by a bulleted list for the sentences below. Something like:

To learn more about the topics discussed in this guide, see the following pages in the {+mdb-server+} manual:
- To view a full list of expression operators, see :manual:`Aggregation Operators </reference/operator/aggregation/>`
...
...
...

jordan-smith721 (approved):
Nice work! LGTM w/ a couple suggestions!

source/includes/aggregation.php
```diff
@@ -0,0 +1,43 @@
...
+$command = [
+    'explain' => [
+        'aggregate' => 'restaurants',
+        'pipeline' => $pipeline,
+        'cursor' => new stdClass()
+    ]
+];
+
+$result = $db->command($command)->toArray();
+echo json_encode($result[0]) . PHP_EOL;
```
jmikola, line 41: In contrast to https://github.com/mongodb/docs-php-library/pull/113, I think you're fine to use `json_encode()` here because the aggregation result documents will only include strings and integers. If that ever changes, you'd want to use PHPC's BSON API to output Extended JSON.

source/includes/aggregation.php
```diff
@@ -0,0 +1,43 @@
+<?php
+require 'vendor/autoload.php';
+
+$uri = getenv('MONGODB_URI') ?: throw new RuntimeException('Set the MONGODB_URI variable to your Atlas URI that connects to the sample dataset');
+$client = new MongoDB\Client($uri);
+
+$db = $client->sample_restaurants;
```
jmikola, line 7: If you end up using `Collection::explain()`, there will be no need to assign a separate `$db` variable.

source/includes/aggregation.php
```diff
@@ -0,0 +1,43 @@
...
+$collection = $db->restaurants;
+
+// Retrieves documents with a cuisine value of "Bakery", groups them by "borough", and
+// counts each borough's matching documents
+// start-match-group
+$pipeline = [
+    ['$match' => ['cuisine' => 'Bakery']],
+    ['$group' => ['_id' => '$borough', 'count' => ['$sum' => 1]]]
```
jmikola, line 15:
```suggestion
    ['$group' => ['_id' => '$borough', 'count' => ['$sum' => 1]]],
```

Note: it's typically good practice to use trailing commas after all array elements, even though it's not required for the final element. I believe we actually enforce this in the PHPLIB coding standard, as it leads to more concise diffs when changing arrays.

This can apply to the `$pipeline` assignment below, too. I assume that is intentionally duplicated since it's displayed in a different location within the tutorial.

source/includes/aggregation.php
```diff
@@ -0,0 +1,43 @@
...
+];
+
+$command = [
+    'explain' => [
+        'aggregate' => 'restaurants',
+        'pipeline' => $pipeline,
+        'cursor' => new stdClass()
+    ]
+];
```
jmikola, line 38: Consider the following to utilize [`MongoDB\Collection::explain()`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBCollection-explain/):

```php
$aggregate = new MongoDB\Operation\Aggregate(
    $collection->getDatabaseName(),
    $collection->getCollectionName(),
    $pipeline
);

$result = $collection->explain($aggregate);
```

Note that `Collection::explain()` returns a single document instead of a cursor, so there's no need to call `toArray()`.

source/aggregation.txt
```diff
@@ -0,0 +1,205 @@
...
+  :manual:`Aggregation Pipeline </core/aggregation-pipeline/>`.
+
+- To learn more about creating pipeline stages, see :manual:`Aggregation
+  Stages </reference/operator/aggregation-pipeline/>`.
+
+- To learn more about explaining MongoDB operations, see
+  :manual:`Explain Output </reference/explain-results/>` and
+  :manual:`Query Plans </core/query-plans/>`.
+
+.. TODO:
```
jmikola, line 193: Does the `php-aggregation-tutorials-landing` target exist already? I'm not sure how `.. TODO:` is interpreted here, and whether that prevents the subsequent content from rendering.

source/aggregation.txt
```diff
@@ -0,0 +1,205 @@
...
+ Aggregation Tutorials
+ ~~~~~~~~~~~~~~~~~~~~~
+
+.. To view step-by-step explanations of common aggregation tasks, see
+.. :ref:`php-aggregation-tutorials-landing`.
+
+API Documentation
+~~~~~~~~~~~~~~~~~
+
+For more information about executing aggregation operations by using the {+php-library+},
```
jmikola, line 203: Would "with the" read better than "by using the"? There's already an active verb with "executing".

source/aggregation.txt
```diff
@@ -0,0 +1,205 @@
...
+      :visible: false
+
+      {"_id":"Brooklyn","count":173}
+      {"_id":"Queens","count":204}
+      {"_id":"Bronx","count":71}
+      {"_id":"Staten Island","count":20}
+      {"_id":"Missing","count":2}
+      {"_id":"Manhattan","count":221}
+
+Explain an Aggregation
```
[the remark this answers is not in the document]
norareidy: No specific reason, but I also see there's an explain option that you can pass to the aggregate() method - that probably makes the most sense for this code example. And it looks like the output only includes the query plan information, so I'll update the copy.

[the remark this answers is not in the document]
norareidy: Actually will use `MongoDB\Collection::explain()`, based on your comment below

source/aggregation.txt
norareidy: We have style guidelines about with / by using: https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#std-term-using

source/aggregation.txt
norareidy: Not yet, so this section is commented out (and will be uncommented when the aggregation tutorials are created)

source/includes/aggregation.php
```diff
@@ -0,0 +1,40 @@
...
+
+$aggregate = new MongoDB\Operation\Aggregate(
+    $collection->getDatabaseName(),
+    $collection->getCollectionName(),
+    $pipeline
+);
+
+$result = $collection->explain($aggregate);
+echo json_encode($result) . PHP_EOL;
```
jmikola, line 38: Note: I dunno offhand if `explain` output might include more complex BSON types, but since this is essentially debugging output I don't think we need to be pedantic about using the driver's BSON functions to print actual extended JSON.

source/includes/aggregation.php
```diff
@@ -0,0 +1,40 @@
...
+// start-match-group
+$pipeline = [
+    ['$match' => ['cuisine' => 'Bakery']],
+    ['$group' => ['_id' => '$borough', 'count' => ['$sum' => 1]]],
+];
+
+$cursor = $collection->aggregate($pipeline);
+
+foreach ($cursor as $doc) {
+    echo json_encode($doc) . PHP_EOL;
```
jmikola, line 20: `echo` supports any number of comma-delimited arguments, so you could also consider:

```suggestion
    echo json_encode($doc), PHP_EOL;
```

This avoids string concatenation and would be slightly more efficient.

Also applies to `echo` for the explain output. I might have missed this in previous PRs, so up to you if you'd like to address. OK to leave as-is if a bunch of docs pages are already doing `. PHP_EOL`.

jmikola (approved):
Optional small suggestions I missed previously.

source/includes/aggregation.php
norareidy: I'll change to a comma here and add that to the cleanup tasks

PR #117, opened 2024-08-28
DOCSP-41962 - Stable API

mongoKart:
Adapted from the [PyMongo page of the same name](https://www.mongodb.com/docs/languages/python/pymongo-driver/current/connect/stable-api/)

# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-41962
Staging - https://preview-mongodbmongokart.gatsbyjs.io/php-library/docsp-41962-stable-api/connect/stable-api/

## Self-Review Checklist


source/connect/stable-api.txt
```diff
@@ -0,0 +1,117 @@
...
+
+The following table describes the parameters of the ``MongoDB\Driver\ServerApi`` constructor.
+You can use these parameters to customize the behavior of the {+stable-api+}.
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+   :widths: 25,75
+
+   * - Option Name
```
norareidy, line 81: S: Since you described these as parameters in the table intro, I'd change this table header to match

```suggestion
   * - Parameter
```

source/connect/stable-api.txt
```diff
@@ -0,0 +1,117 @@
...
+the commands it supports, see :manual:`Stable API </reference/stable-api/>` in the
+{+mdb-server+} manual.
+
+Enable the {+stable-api+}
+-------------------------
+
+To enable the {+stable-api+}, perform the following steps:
+
+1. Construct a ``MongoDB\Driver\ServerApi`` object and pass the {+stable-api+}
+   version you want to use. Currently, the driver supports only version 1.
```
norareidy, line 47: S: same suggestion as above
```suggestion
   version you want to use. Currently, the library supports only version 1.
```

norareidy (approved):
LGTM with some suggestions

source/includes/connect/stable-api.php
```diff
@@ -0,0 +1,23 @@
+// start-specify-v1
```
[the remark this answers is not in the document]
mongoKart: yeah, the code shows the most straightforward way to do this, but I couldn't explain it in a neat chronology

source/connect/tls.txt
```diff
@@ -265,4 +265,4 @@ API Documentation
 To learn more about configuring TLS for the {+driver-short+},
 see the following API documentation:
 
-- :ref:`MongoDB\Client <php-mongodb-client>`
\ No newline at end of file
+- :ref:`MongoDB\Client <php-api-mongodbclient>`
```
jmikola, line 268: Why does this used a named reference instead of `:phpclass:` or `:phpmethod:`? Those are used throughout existing docs (e.g. [GridFS tutorial](https://github.com/mongodb/docs-php-library/blob/master/source/tutorial/gridfs.txt)).

source/includes/connect/stable-api.php
```diff
@@ -0,0 +1,23 @@
...
+<?php
+
+$uri = "mongodb://<hostname>:<port>";
+
+$driverOptions = [
+    'serverApi' => new MongoDB\Driver\ServerApi('1')
```
jmikola, line 7:
```suggestion
    'serverApi' => new MongoDB\Driver\ServerApi('1'),
```

If you're breaking PHP arrays into multiple lines, it's consistent to include a trailing comma after each element. Alternatively, you can inline the entire array on a single line:

```php
$driverOptions = ['serverApi' => new MongoDB\Driver\ServerApi('1')];
```

source/includes/connect/stable-api.php
```diff
@@ -0,0 +1,23 @@
...
+];
+
+$client = new MongoDB\Client($uri, [], $driverOptions);
+// end-specify-v1
+
+// start-stable-api-options
+<?php
```
jmikola, line 14: I don't recall seeing multiple `<?php` tags in source files from @norareidy's previous PRs. This makes the file itself invalid, unless you intend to add a `?>` above (probably between `end-specify-v1` and `start-stable-api-options` since there's no reason to show `?>` to a user (they won't use it in their files).

If these source files are ever meant to be executed for correctness (as we do for some existing sources in the PHPLIB docs before the docs team took over), that'd be even more reason to ensure the syntax within is valid.

source/connect/stable-api.txt
```diff
@@ -0,0 +1,117 @@
...
+   version. If you need to run commands using more than one version of the 
+   {+stable-api+}, create a new ``MongoDB\Client`` instance.
+
+.. _stable-api-options:
+
+Configure the {+stable-api+}
+------------------------
+
+The following table describes the parameters of the ``MongoDB\Driver\ServerApi`` constructor.
+You can use these parameters to customize the behavior of the {+stable-api+}.
```
jmikola, line 74: The first argument to the [ServerApi constructor](https://www.php.net/manual/en/class.mongodb-driver-serverapi.php) is `$version`, which isn't included in this table. I realize that's demonstrated in the preceding example, but I wonder if this table might mislead readers into thinking `strict` is the first parameter.

This relates to my later comment on the PHP code to suggest using named arguments, which would make the code more self-documenting.

source/includes/connect/stable-api.php
```diff
@@ -0,0 +1,23 @@
...
+
+$uri = "mongodb://<hostname>:<port>";
+
+$driverOptions = [
+    'serverApi' => new MongoDB\Driver\ServerApi('1', true, true)
```
jmikola, line 19: @alcaeus: I'm not sure what version of PHPLIB is being targeted here, but if we're bumping to PHP 8.1+ with version 1.20 I think this would be a good opportunity to use [named arguments](https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments) (available in PHP 8.0+) for the the booleans. Consider:

```
'serverApi' => new MongoDB\Driver\ServerApi('1', strict: true, deprecationErrors: true),
```

Ignore this suggestion if we need to remain compatible with earlier versions (back to PHP 7.4).

Also, my previous comment about a trailing comma after elements in multi-line arrays also applies here.

source/connect/stable-api.txt
```diff
@@ -0,0 +1,117 @@
...
+      :end-before: // end-stable-api-options
+      :emphasize-lines: 5-7, 9
+
+API Documentation
+-----------------
+
+For more information about the ``MongoDB\Client`` class, see the following {+driver-short+}
+API documentation:
+
+- :ref:`MongoDB\Client <php-api-mongodbclient>`
```
jmikola, line 112: Same comment below about using `:phpclass:` instead. I'm actually not sure why it's necessary to reference this in a bullet instead of just hyperlinking ``MongoDB\Client``  in the above paragraph.

source/connect/stable-api.txt
```diff
@@ -0,0 +1,117 @@
...
+   :header-rows: 1
+   :stub-columns: 1
+   :widths: 25,75
+
+   * - Parameter
+     - Description
+
+   * - strict
+     - | **Optional**. When ``true``, if you call a command that isn't part of 
+         the declared API version, the library raises an exception.
```
jmikola, line 86: Server API errors are enforced by the server. This is relevant to clarify if users would want to know whether to expect a client-side exception or something derived from [MongoDB\Driver\Exception\ServerException](https://www.php.net/manual/en/class.mongodb-driver-serverdescription.php).

I believe this point also came up in one of @norareidy's previous PRs, although I don't recall whether we decided to leave that as-is or clarify in the merged text.

This applies to `deprecationErrors` below as well.

source/connect/stable-api.txt
```diff
@@ -0,0 +1,117 @@
...
+       |
+       | Default: ``false``
```
jmikola, line 88: I imagine you don't care to clarify this here, but the PHP extension actually defaults to `null`. It's the server that applies a `false` default (per [Stable API docs](https://www.mongodb.com/docs/manual/reference/stable-api/#parameters)).

I only bring this up as the [extension docs](https://www.php.net/manual/en/mongodb-driver-serverapi.construct.php) do correctly state:

> If no value is provided, the server default value (`false`) is used.

This applies to `deprecationErrors` below as well.

source/includes/connect/stable-api.php
alcaeus: I think we can assume newer PHP versions, so using named arguments sounds like a great idea, especially for those boolean flags!

source/connect/tls.txt
mongoKart: I didn't know about these directives (and apparently the other writers didn't either). We've agreed to use them for API links.

source/connect/stable-api.txt
mongoKart: reworded

source/connect/stable-api.txt
mongoKart: reworded

source/connect/stable-api.txt
mongoKart: that's the way we link to api docs on all pages, probably because many of them include multiple links.

source/includes/connect/stable-api.php
mongoKart: i wasn't sure about these. the existing, un-standardized docs (e.g., hhttps://www.mongodb.com/docs/php-library/current/tutorial/gridfs/#uploading-files-with-writable-streamsere) include them in all code samples, so i assumed it was the preferred syntax.

these files will never be executed _in toto_, so i'm just going to remove these tags.

source/includes/connect/stable-api.php
mongoKart: love it. helps a ton with readability.

source/connect/stable-api.txt
mongoKart: Nora said she corrected it the server throwing an exception, which is what I've done here too.

source/includes/connect/stable-api.php
```diff
@@ -0,0 +1,13 @@
...
+$uri = "mongodb://<hostname>:<port>";
+
+$driverOptions = ['serverApi' => new MongoDB\Driver\ServerApi('1')];
+$client = new MongoDB\Client($uri, [], $driverOptions);
+// end-specify-v1
+
+// start-stable-api-options
+$uri = "mongodb://<hostname>:<port>";
+
+$driverOptions = ['serverApi' => new MongoDB\Driver\ServerApi('1', strict: true, deprecationErrors: true)];
```
mongoKart, line 11: If I wanted to split this line into multiple, shorter lines, how would I do that?

source/includes/connect/stable-api.php
jmikola: > include them in all code samples

I think that was necessary for syntax highlighting. Note that most of those code blocks are embedded in the RST and weren't being included from a common file (nor do they get automated testing).

Speaking just of the [GridFS examples](https://www.mongodb.com/docs/php-library/current/tutorial/gridfs/), I'll also note that those are all full scripts that can be executed on their own (each constructs a Client).

I think it makes sense to leave `<?php` at the top of each of your files and won't hurt to show it in the lead-in example that constructs a client (or the sample app), but if you don't need it for syntax highlighting feel free to get rid of it.

source/includes/connect/stable-api.php
```diff
@@ -0,0 +1,13 @@
+// start-specify-v1
```
jmikola, line 1:
```suggestion
<?php

// start-specify-v1
```

Adding back the opening PHP tag so the file itself is valid PHP code. The only problem with the previous state of this file was the duplicate `<?php` tag.

source/connect/stable-api.txt
```diff
@@ -0,0 +1,119 @@
...
+
+Enable the {+stable-api+}
+-------------------------
+
+To enable the {+stable-api+}, perform the following steps:
+
+1. Construct a ``MongoDB\Driver\ServerApi`` object and pass the {+stable-api+}
+   version you want to use. Currently, the library supports only version 1.
+#. Construct a ``MongoDB\Client`` object. For the ``driverOptions`` parameter, pass an
+   array that contains the ``serverApi`` connection option. Set this option to the
```
jmikola, line 49:
```suggestion
   array that contains the ``serverApi`` option. Set this option to the
```

To avoid confusion, I wouldn't refer to this as a "connection option". When constructing a Client (or Manager in the extension directly), we refer to either URI options (specified in either the connection string or second array argument) or driver options (the third array param).

source/includes/connect/stable-api.php
jmikola: I think this would be readable:

```php
$uri = "mongodb://<hostname>:<port>";

$serverApi = new MongoDB\Driver\ServerApi('1', strict: true, deprecationErrors: true);
$driverOptions = ['serverApi' => $serverApi];

$client = new MongoDB\Client($uri, [], $driverOptions);
```

I also think the previous example could benefit from an extra newline before `$client = ...`.

----

If you want to break ServerApi across multiple lines, you can do something like:

```php
$serverApi = new MongoDB\Driver\ServerApi(
    '1',
    strict: true,
    deprecationErrors: true,
);
```

PHP 8.0 allows trailing commas in parameter lists (consistent with what we'd do for multi-line arrays): https://php.watch/versions/8.0/trailing-comma-parameter-use-list

If that's too much wrapping, you could also do this:

```php
$driverOptions = [
    'serverApi' => new MongoDB\Driver\ServerApi('1', strict: true, deprecationErrors: true),
];
```

That said, I don't think either of these are better than just assigning `$serverApi` on its own line as I did in the first suggestion.

source/includes/connect/stable-api.php
```diff
@@ -0,0 +1,13 @@
...
+$uri = "mongodb://<hostname>:<port>";
+
+$driverOptions = ['serverApi' => new MongoDB\Driver\ServerApi('1')];
+$client = new MongoDB\Client($uri, [], $driverOptions);
```
jmikola, line 5:
```suggestion

$client = new MongoDB\Client($uri, [], $driverOptions);
```

Consider for readability.

jmikola (approved):

PR #127, opened 2024-09-05
DOCSP-41983: indexes landing pg

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-41983>
Staging - https://preview-mongodbrustagir.gatsbyjs.io/php-library/DOCSP-41983-indexes-landing/indexes/

## Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST? **not related to pr**

### DIFF
--- a/source/includes/usage-examples/index-code-examples.php
+++ b/source/includes/usage-examples/index-code-examples.php
@@ -1,0 +1,79 @@
+<?php
+
+require __DIR__ . '/../vendor/autoload.php';
+
+use MongoDB\Client;
+
+$client = new Client('<connection string>');
+$collection = $client-><database>-><collection>;
+
+// start-single-field
+$result = $collection->createIndex(['<field name>' => 1]);
+// end-single-field
+
+// start-compound
+$indexName = $collection->createIndex(
+    ['<field name 1>' => 1, '<field name 2>' => 1]
+);
+// end-compound
+
+// start-multikey
+$result = $collection->createIndex(['<array field name>' => 1]);
+// end-multikey
+
+// start-search-create
+$result = $collection->createSearchIndex(
+    ['mappings' => ['dynamic' => true]],
+    ['name' => '<index name>']
+);
+// end-search-create
+
+// start-search-list
+foreach ($collection->listIndexes() as $indexInfo) {
+    var_dump($indexInfo);
+}
+// end-search-list
+
+// start-search-update
+$result = $collection->updateSearchIndex(
+    ['name' => '<index name>'],
+    ['mappings' => ['dynamic' => true]],
+ );
+// end-search-update
+
+// start-search-delete
+$result = $collection->dropIndex('<index name>');
+// end-search-delete
+
+// start-text
+$result = $collection->createIndex(['<field name>' => 'text']);
+// end-text
+
+// start-geo
+$result = $collection->createIndex(
+    [ '<GeoJSON object field>' => '2dsphere'], ['name' => '<index name>']
+);
+// end-geo
+
+// start-unique
+$result = $collection->createIndex(['<field name>' => 1], ['unique' => true]);
+// end-unique
+
+// start-wildcard
+$result = $collection->createIndex(['$**' => 1]);
+// end-wildcard
+
+// start-clustered
+$options = [
+    'clusteredIndex' => [
+        'key' => ['_id' => 1],
+        'unique' => true
+    ]
+];
+
+$database->createCollection('<collection name>', $options);
+// end-clustered
+
+// start-remove
+$result = $collection->dropIndex('<index name>');
+// end-remove
--- a/source/includes/usage-examples/index-code-examples.php
+++ b/source/includes/usage-examples/index-code-examples.php
@@ -38,7 +38,7 @@ $indexName = $collection->createSearchIndex(
 
 // start-search-list
 foreach ($collection->listSearchIndexes() as $indexInfo) {
-    echo toJSON($indexInfo) . PHP_EOL;
+    echo toJSON($indexInfo) , PHP_EOL;
 }
 // end-search-list
 
--- a/source/includes/usage-examples/index-code-examples.php
+++ b/source/includes/usage-examples/index-code-examples.php
@@ -92,7 +92,7 @@ $database->createCollection('<collection name>', $options);
 
 // start-list
 foreach ($collection->listIndexes() as $indexInfo) {
-    echo toJSON($indexInfo) . PHP_EOL;
+    echo $indexInfo;
 }
 // end-list
 
--- a/source/includes/usage-examples/index-code-examples.php
+++ b/source/includes/usage-examples/index-code-examples.php
@@ -29,7 +29,7 @@ $result = $collection->createSearchIndex(
 // end-search-create
 
 // start-search-list
-foreach ($collection->listIndexes() as $indexInfo) {
+foreach ($collection->listSearchIndexes() as $indexInfo) {
     var_dump($indexInfo);
 }
 // end-search-list
@@ -73,6 +73,12 @@ $options = [
 
 $database->createCollection('<collection name>', $options);
 // end-clustered
+
+// start-list
+foreach ($collection->listIndexes() as $indexInfo) {
+    var_dump($indexInfo);
+}
+// end-list
 
 // start-remove
 $result = $collection->dropIndex('<index name>');
--- a/source/includes/usage-examples/index-code-examples.php
+++ b/source/includes/usage-examples/index-code-examples.php
@@ -2,13 +2,21 @@
 
 require __DIR__ . '/../vendor/autoload.php';
 
-use MongoDB\Client;
+$uri = getenv('MONGODB_URI') ?: throw new RuntimeException('Set the MONGODB_URI variable to your Atlas URI that connects to the sample dataset');
+$client = new MongoDB\Client($uri);
 
-$client = new Client('<connection string>');
-$collection = $client-><database>-><collection>;
+$database = $client->sample_db;
+$collection = $database->sample_coll;
+
+// start-to-json
+function toJSON(object $document): string
+{
+    return MongoDB\BSON\Document::fromPHP($document)->toRelaxedExtendedJSON();
+}
+// end-to-json
 
 // start-single-field
-$result = $collection->createIndex(['<field name>' => 1]);
+$indexName = $collection->createIndex(['<field name>' => 1]);
 // end-single-field
 
 // start-compound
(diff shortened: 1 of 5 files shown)


source/indexes.txt
```diff
@@ -0,0 +1,274 @@
...
+.. literalinclude:: /includes/usage-examples/sample-app.php
+   :language: php
+   :copyable:
+   :linenos:
+   :emphasize-lines: 10-12
+
+Single Field Index
+------------------
+
+The following example creates an ascending index on the specified field:
```
mcmorisi, line 58: I noticed that this is the only example introduction that specifies ascending index – could be a good idea to ensure that it's called out for all other examples that create an ascending index.

mcmorisi (approved):
LGTM with one suggestion!

source/includes/usage-examples/index-code-examples.php
```diff
@@ -0,0 +1,85 @@
...
+// end-clustered
+
+// start-list
+foreach ($collection->listIndexes() as $indexInfo) {
+    var_dump($indexInfo);
+}
+// end-list
+
+// start-remove
+$result = $collection->dropIndex('<index name>');
```
jmikola, line 84: [`dropIndex()`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBCollection-dropIndex/#return-values) returns a command result document, which has little value to the user in this case. I'd suggest omitting the assignment here:

```suggestion
$collection->dropIndex('<index name>');
```

source/indexes.txt
```diff
@@ -0,0 +1,278 @@
...
+   :dedent:
+
+.. TODO To learn more about multikey indexes, see the :ref:`php-multikey-index`
+.. guide.
+
+Geospatial Index
+----------------
+
+The following example creates a 2dsphere index on the specified field
+that has GeoJSON object values:
```
jmikola, line 105: Consider linking GeoJSON to https://www.mongodb.com/docs/manual/reference/geojson/.

I assume that page may be linked indirectly by whatever geospatial docs you intend to add for the TODO below, but a direct link here would help readers that aren't familiar with GeoJSON (much less geospatial indexes).

source/indexes.txt
```diff
@@ -0,0 +1,278 @@
...
+   :language: php
+   :copyable:
+   :dedent:
+
+.. To learn more about compound indexes, see the :ref:`php-compound-index` guide.
+
+Multikey Index
+--------------
+
+The following example creates an ascending multikey index on the
```
jmikola, line 88:
```suggestion
The following example creates an ascending, multikey index on the
```

Does the style guide call for commas between sequential adjectives? If so, this suggestion applies to some later index descriptions in this file as well.

source/indexes.txt
```diff
@@ -0,0 +1,278 @@
...
+   :dedent:
+
+.. TODO: To learn more about clustered indexes, see the :ref:`php-clustered-index`
+.. guide.
+
+Atlas Search Index Management
+-----------------------------
+
+The following sections contain code examples that describe how to manage
+:atlas:`Atlas Search indexes </atlas-search/manage-indexes/>`.
```
jmikola, line 170: Not sure if this is better addressed in some common template for driver docs on Atlas Search Index Management, but this seems like a good place to note that all of the driver APIs for search indexes are async. The driver methods themselves may be sync, but the main difference is that the server returns its response immediately while delegating the index management to happen in the background. This is why some of the existing driver example code elsewhere uses `listSearchIndexes()` to determine when the index is actually created and ready to use after initially invoking `createSearchIndex()` (or `createSearchIndexes()` in the case of a batch creation).

source/indexes.txt
```diff
@@ -0,0 +1,278 @@
...
+   :start-after: start-search-delete
+   :end-before: end-search-delete
+   :language: php
+   :copyable:
+   :dedent:
+
+.. To learn more about deleting search indexes, see the :ref:`php-atlas-search-index-drop`
+.. guide.
+
+Text Index
```
jmikola, line 237: I find it odd that the Atlas Search Index Management section appears in the middle of this page and is flanked by other sections that apply to non-search indexes. IMO, it'd be more clear to the reader to segregate all search and non-search index content. This can either be done with a separate heading for the non-search indexes, or just by putting Atlas Search Index Management at the very end.

This wasn't so much an issue with Text Index as it was with the List and Delete sections that follow. It seemed inconsistent to not present List and Delete after the Create examples, as was the natural case within the Atlas Search Index Management section.

source/includes/usage-examples/sample-app.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+require __DIR__ . '/../vendor/autoload.php';
+
+use MongoDB\Client;
+
+$client = new Client('<connection string>');
+$collection = $client-><database>-><collection>;
```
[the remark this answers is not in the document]
rustagir: The preceding text that introduces the sample app gives instructions to users on adding their connection string! I will update the code to using strings to represent placeholders for db and collection.

source/indexes.txt
```diff
@@ -0,0 +1,278 @@
...
+:ref:`sample application <php-index-sample>` or your own application.
+Be sure to replace all placeholders in the code examples, such as
+``<connection string>``, with the relevant values for your MongoDB
+deployment.
+
+.. _php-index-sample:
+
+.. include:: /includes/usage-examples/sample-app-intro.rst
+
+.. literalinclude:: /includes/usage-examples/sample-app.php
```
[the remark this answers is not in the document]
rustagir: will adapt the code to be more similar to the other files

source/includes/usage-examples/index-code-examples.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+require __DIR__ . '/../vendor/autoload.php';
+
+use MongoDB\Client;
+
+$client = new Client('<connection string>');
```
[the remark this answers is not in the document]
rustagir: Changed

source/includes/usage-examples/index-code-examples.php
```diff
@@ -0,0 +1,85 @@
...
+
+// start-search-list
+foreach ($collection->listSearchIndexes() as $indexInfo) {
+    var_dump($indexInfo);
+}
+// end-search-list
+
+// start-search-update
+$result = $collection->updateSearchIndex(
+    ['name' => '<index name>'],
```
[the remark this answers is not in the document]
rustagir: my bad - no I did not test the AS index code examples, I pulled them from the API docs.

source/includes/usage-examples/index-code-examples.php
```diff
@@ -0,0 +1,85 @@
...
+    ['mappings' => ['dynamic' => true]],
```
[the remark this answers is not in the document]
rustagir: I'll implement a more useful example

source/includes/usage-examples/index-code-examples.php
```diff
@@ -0,0 +1,85 @@
...
+$result = $collection->dropIndex('<index name>');
+// end-search-delete
+
+// start-text
+$result = $collection->createIndex(['<field name>' => 'text']);
+// end-text
+
+// start-geo
+$result = $collection->createIndex(
+    [ '<GeoJSON object field>' => '2dsphere'], ['name' => '<index name>']
```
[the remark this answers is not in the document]
rustagir: removing that param

source/includes/usage-examples/index-code-examples.php
```diff
@@ -0,0 +1,85 @@
...
+
+// start-clustered
+$options = [
+    'clusteredIndex' => [
+        'key' => ['_id' => 1],
+        'unique' => true
+    ]
+];
+
+$database->createCollection('<collection name>', $options);
```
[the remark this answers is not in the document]
rustagir: fixed the db/coll handling

source/indexes.txt
rustagir: i'll add a note! but yes, this can be a point made more clear in a dedicated page

source/includes/usage-examples/index-code-examples.php
```diff
@@ -0,0 +1,101 @@
+<?php
+
+require __DIR__ . '/../vendor/autoload.php';
+
+$uri = getenv('MONGODB_URI') ?: throw new RuntimeException('Set the MONGODB_URI variable to your Atlas URI that connects to the sample dataset');
+$client = new MongoDB\Client($uri);
+
+$database = $client->sample_db;
+$collection = $database->sample_coll;
```
jmikola, line 9: It looks like `sample-app.php` was changed to use:

```php
$collection = $client->selectCollection('<database>', '<collection>');
```

Did you mean to do the same here for consistency? It doesn't look like this content is presented to the user at all since it's outside of start/end comments.

source/includes/usage-examples/index-code-examples.php
```diff
@@ -0,0 +1,101 @@
...
+// start-search-create
+$indexName = $collection->createSearchIndex(
+    ['mappings' => ['dynamic' => true]],
+    ['name' => '<Search index name>']
+);
+// end-search-create
+
+// start-search-list
+foreach ($collection->listSearchIndexes() as $indexInfo) {
+    echo toJSON($indexInfo) . PHP_EOL;
```
jmikola, line 41:
```suggestion
    echo toJSON($indexInfo), PHP_EOL;
```

See: https://github.com/mongodb/docs-php-library/pull/115#discussion_r1750423119

source/includes/usage-examples/index-code-examples.php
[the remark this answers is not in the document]
jmikola: You might be able to pull some inspiration from examples I wrote up in the Doctrine ODM project: https://github.com/doctrine/mongodb-odm/commit/fed919fac08b965e0741f8d6f3c93b07ca0a46e7

In particular, there is some code in [`AbstractMappingDriverTestCase.php`](https://github.com/doctrine/mongodb-odm/commit/fed919fac08b965e0741f8d6f3c93b07ca0a46e7#diff-21aa2b70881e07fae85b672c300e17afd8eb4681c81a60fe635d518a698f6a10) that presents both dynamic and non-dynamic search index definitions. Those are expected values from parsing annotations and XML mappings, which are probably less helpful.

I think some of that content is also available via: https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/search-indexes.html

source/includes/usage-examples/index-code-examples.php
rustagir: This is not visible to users, so it didn't seem necessary to update

rustagir:
@jmikola I can take those examples under consideration for the longer Atlas Search index focused guide that is planned: https://jira.mongodb.org/browse/DOCSP-41987

jmikola (approved):

PR #128, opened 2024-09-06
DOCSP-41970: Delete documents

norareidy:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-41970
Staging - https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-41970-delete/write/delete/

## Self-Review Checklist

- [ ] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?

### DIFF
--- a/source/includes/write/delete.php
+++ b/source/includes/write/delete.php
@@ -1,0 +1,35 @@
+<?php
+require 'vendor/autoload.php'; // Ensure you have MongoDB PHP library installed
+
+$uri = getenv('MONGODB_URI') ?: throw new RuntimeException('Set the MONGODB_URI variable to your Atlas URI that connects to the sample dataset');
+$client = new MongoDB\Client($uri);
+
+// start-db-coll
+$collection = $client->sample_restaurants->restaurants;
+// end-db-coll
+
+// Deletes a document that has a "name" value of "Ready Penny Inn"
+// start-delete-one
+$result = $collection->deleteOne(['name' => 'Ready Penny Inn']);
+// end-delete-one
+
+// Deletes documents that have a "borough" value of "Brooklyn"
+// start-delete-many
+$result = $collection->deleteMany(['borough' => 'Brooklyn']);
+// end-delete-many
+
+// Deletes matching documents and attaches a comment to the operation
+// start-delete-options
+$result = $collection->deleteMany(
+    ['name' => new MongoDB\BSON\Regex('Mongo', '')],
+    ['comment' => 'Deleting Mongo restaurants'],
+);
+// end-delete-options
+
+// Deletes and prints the number of documents that have a "cuisine" value of "Greek"
+// start-delete-count
+$result = $collection->deleteMany(['cuisine' => 'Greek']);
+echo $result->getDeletedCount() . PHP_EOL;
+// end-delete-count
+
+?>
--- a/source/includes/write/delete.php
+++ b/source/includes/write/delete.php
@@ -29,6 +29,6 @@ $result = $collection->deleteMany(
 // Deletes and prints the number of documents that have a "cuisine" value of "Greek"
 // start-delete-count
 $result = $collection->deleteMany(['cuisine' => 'Greek']);
-echo 'Deleted documents: ' . $result->getDeletedCount() . PHP_EOL;
+echo 'Deleted documents: ' , $result->getDeletedCount() , PHP_EOL;
 // end-delete-count
 
--- a/source/includes/write/delete.php
+++ b/source/includes/write/delete.php
@@ -10,18 +10,18 @@ $collection = $client->sample_restaurants->restaurants;
 
 // Deletes a document that has a "name" value of "Ready Penny Inn"
 // start-delete-one
-$result = $collection->deleteOne(['name' => 'Ready Penny Inn']);
+$collection->deleteOne(['name' => 'Ready Penny Inn']);
 // end-delete-one
 
 // Deletes documents that have a "borough" value of "Brooklyn"
 // start-delete-many
-$result = $collection->deleteMany(['borough' => 'Brooklyn']);
+$collection->deleteMany(['borough' => 'Brooklyn']);
 // end-delete-many
 
 // Deletes matching documents and attaches a comment to the operation
 // start-delete-options
-$result = $collection->deleteMany(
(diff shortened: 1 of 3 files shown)


mcmorisi (approved):
LGTM, nice work!

source/includes/write/delete.php
```diff
@@ -0,0 +1,34 @@
...
+
+// Deletes documents that have a "borough" value of "Brooklyn"
+// start-delete-many
+$result = $collection->deleteMany(['borough' => 'Brooklyn']);
+// end-delete-many
+
+// Deletes matching documents and attaches a comment to the operation
+// start-delete-options
+$result = $collection->deleteMany(
+    ['name' => new MongoDB\BSON\Regex('Mongo', '')],
```
jmikola, line 24: If you're not using an options (i.e. regex flags), you can omit the optional second parameter:

```suggestion
    ['name' => new MongoDB\BSON\Regex('Mongo')],
```

----

Do other language docs use the BSON regex type or the [`$regex`](https://www.mongodb.com/docs/manual/reference/operator/query/regex/) query operator? If we want to ensure consistency across all language examples, the operator might be the way to go.

The server docs [note differences](https://www.mongodb.com/docs/manual/reference/operator/query/regex/#-regex-vs.--pattern--syntax) and point out where you'd need the BSON type; however, they're also written from the perspective of the shell where Javascript's [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) (object or literal) is used and then serialized to a BSON type over the wire. RegExp has some of its own limitations (as do native regex types in other languages). In contrast, PHP has no native regex object and the `MongoDB\BSON\Regex` object is just a container for two strings (pattern and options), which serialize as-is to a BSON regex type (i.e. we can express anything the [server supports](https://bsonspec.org/spec.html)).

source/write/delete.txt
```diff
@@ -0,0 +1,197 @@
...
+   :start-after: start-delete-many
+   :end-before: end-delete-many
+   :language: php
+   :dedent:
+
+Modify the Delete Operation
+---------------------------
+
+You can modify the behavior of the ``MongoDB\Collection::deleteOne()`` and
+``MongoDB\Collection::deleteMany()`` methods by by passing an array that
```
jmikola, line 91:
```suggestion
``MongoDB\Collection::deleteMany()`` methods by passing an array that
```

source/write/delete.txt
```diff
@@ -0,0 +1,197 @@
...
+
+.. list-table::
+   :widths: 30 70
+   :header-rows: 1
+
+   * - Option
+     - Description
+
+   * - ``collation``
+     - | Specifies the kind of language collation to use when sorting
```
jmikola, line 103: Is "sorting" the correct term to use here? For a delete operation I'd expect `collation` to only impact matching.

Even for a `find` command, I expect `collation` would be used for matching _and_ sorting.

source/write/delete.txt
```diff
@@ -0,0 +1,197 @@
...
+Return Value
+------------
+
+The ``MongoDB\Collection::deleteOne()`` and ``MongoDB\Collection::deleteMany()``
+methods return a ``MongoDB\DeleteResult`` object. This class contains the
+following member functions:
+
+- ``isAcknowledged()``, which returns a boolean indicating whether the operation
+  was acknowledged
+- ``getDeletedCount()``, which returns the number of documents deleted
```
jmikola, line 163: I'm not sure if this is something you want to discuss here, but the getter methods throw if the write was not acknowledged, since the server wouldn't have returned any information and thus the driver has nothing to report. An exception (no pun intended) to this rule is getter methods for inserted IDs (on `insertOne()` and `insertMany()`) can always be called since those are collected client-side before we talk to the server.

Generally, most users don't run unacknowledged writes. And even if they did, there's probably little reason they'd call `isAcknowledged()` since they'd already know how they're executing the delete operation.

Probably TMI, but just wanted to provide context on how the methods on these result objects relate to each other. This isn't really discussed elsewhere beyond the API docs (e.g. [`DeleteResult::getDeletedCount()`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBDeleteResult-getDeletedCount/).

source/write/delete.txt
norareidy: thanks for catching!

source/write/delete.txt
norareidy: Changed to "comparing strings," since I think that better describes what happens during matching and sorting

source/write/delete.txt
```diff
@@ -0,0 +1,197 @@
...
+
+   * - ``collation``
+     - | Specifies the kind of language collation to use when sorting
+         results. For more information, see :manual:`Collation </reference/collation/#std-label-collation>`
+         in the {+mdb-server+} manual.
+
+   * - ``writeConcern``
+     - | Sets the write concern for the operation.
+         For more information, see :manual:`Write Concern </reference/write-concern/>`
+         in the {+mdb-server+} manual.
```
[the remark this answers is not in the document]
norareidy: I added a note here that the write concern can be inherited, since we do that in other docs. For collation, it seems like potentially too specific

source/write/delete.txt
norareidy: Added that the getDeletedCount() method throws an error if the write was not acknowledged.

jmikola (approved):

source/includes/write/delete.php
norareidy: Good point, it's kind of a mix - the [PyMongo](https://www.mongodb.com/docs/languages/python/pymongo-driver/current/write/delete/#customize-the-delete-operation) and [Go](https://www.mongodb.com/docs/drivers/go/current/fundamentals/crud/read-operations/query-document/#example-4) docs use $regex, and the [Java sync](https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/crud/query-document/#evaluation-operators) and [Reactive Streams](https://www.mongodb.com/docs/languages/java/reactive-streams-driver/upcoming/write/write-delete-documents/#example) docs use helper methods, for ex. So it wouldn't really help with consistency to change this example to use the $regex operator.

PR #129, opened 2024-09-06
DOCSP-41956: run a command

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-41956>
Staging - https://deploy-preview-129--docs-php-library.netlify.app/run-command/

## Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST? **orphan page**


source/write/run-command.txt
```diff
@@ -0,0 +1,182 @@
...
+Execute a Command
+-----------------
+
+To run a database command, you must specify the command and any relevant
+parameters in a command document, then pass the command document to the
+``command()`` method. This method returns a ``Cursor`` object.
+
+The following code shows how you can use the ``command()``
+method to run the ``hello`` command, which returns information about
+the current member's role in the replica set, on a database:
```
jordan-smith721, line 51: The code example is showing the `ping` command. Update the description or update the code example to run 'hello' instead. (`ping` description in suggestion taken from the server manual)
```suggestion
The following code shows how you can use the ``command()``
method to run the ``ping`` command, which is a no-op command used to test whether a server is responding:
```

source/write/run-command.txt
```diff
@@ -0,0 +1,182 @@
...
+   .. code-block:: php
+
+      $readPref = new MongoDB\Driver\ReadPreference('primaryPreferred');
+      $result = $database->command(
+          ['ping' => 1], 
+          ['readPreference' => $readPref]
+      );
+
+   For more information on read preference options, see :manual:`Read
+   Preference </core/read-preference/>` in the Server manual.
```
jordan-smith721, line 80: S: to align with our other repos
```suggestion
   Preference </core/read-preference/>` in the MongoDB Server manual.
```

source/write/run-command.txt
```diff
@@ -0,0 +1,182 @@
...
+       or failed (``0``).
+
+   * - ``operationTime``
+     - The logical time of the operation. MongoDB uses the
+       logical time to order operations.
+       
+       .. seealso::
+  
+          To learn more about logical time, see our blog post about the
+          :website:`Global Logical Clock </blog/post/transactions-background-part-4-the-global-logical-clock>`.
```
jordan-smith721, line 117: Take this out of an admonition since we are moving towards avoiding nested components

source/write/run-command.txt
```diff
@@ -0,0 +1,182 @@
...
+         of the key used to sign the cluster time
+
+.. _rust-command-example:
+
+Command Example
+---------------
+
+The following code shows how you can use the ``command()`` method to run
+the ``explain`` command for a ``count`` operation on the ``flowers``
+collection of the ``plants`` database. The ``explain`` command runs in the
```
jordan-smith721, line 136: S: Simplify
```suggestion
The following example uses the ``command()`` method to run
the ``explain`` command for a ``count`` operation. The ``explain`` command runs in the
```

source/write/run-command.txt
```diff
@@ -0,0 +1,182 @@
...
+``"queryPlanner"`` verbosity mode:
+
+.. literalinclude:: /includes/write/run-command.php
+   :language: php
+   :dedent:
+   :start-after: start-runcommand
+   :end-before: end-runcommand
+
+Output
+~~~~~~
```
jordan-smith721, line 146: Nit: I think we could probably remove this header, I don't think it adds much value

source/write/run-command.txt
```diff
@@ -0,0 +1,182 @@
...
+   "gitVersion":"..."},"serverParameters":{...},"ok":1,"$clusterTime":{...},"signature":{...},
+   "keyId":...}},"operationTime":{"$timestamp":{"t":1725637892,"i":12}}}]"
+
+.. _php-addtl-info-runcommand:
+
+Additional Information
+----------------------
+
+For more information about the concepts in this guide, see the following
+documentation in the Server manual:
```
jordan-smith721, line 172:
```suggestion
documentation in the MongoDB Server manual:
```

source/write/run-command.txt
```diff
@@ -0,0 +1,182 @@
...
+
+- :manual:`db.runCommand() </reference/method/db.runCommand/>`
+- :manual:`Database Commands </reference/command/>`
+- :manual:`hello Command </reference/command/hello/>`
```
jordan-smith721, line 176: If the earlier code example stays the same
```suggestion
- :manual:`ping Command </reference/command/ping/>`
```

jordan-smith721 (changes requested):
Nice job! Just a few suggestions

source/write/run-command.txt
rustagir: oops let me rework!

jordan-smith721 (approved):
LGTM!

source/write/run-command.txt
```diff
@@ -0,0 +1,185 @@
...
+
+.. _php-execute-command:
+
+Execute a Command
+-----------------
+
+To run a database command, you must specify the command and any relevant
+parameters in a command document, then pass the command document to the
+``MongoDB\Database::command()`` method. This method returns a
+``Cursor`` object.
```
GromNaN, line 48: Should you add a link to https://www.php.net/manual/en/class.mongodb-driver-cursor.php?

source/write/run-command.txt
```diff
@@ -0,0 +1,185 @@
...
+
+The following code shows how you can use the ``command()``
+method on a database to run the ``hello`` command, which returns
+information about the current member's role in the replica set:
+
+.. code-block:: php
+
+   $result = $database->command(['hello' => 1]);
```
GromNaN, line 56: You don't explain what `$database` is. It is an instance of `MongoDB\Driver\Database`  obtained by calling `$client->selectDatabase(<name>)`.

source/write/run-command.txt
```diff
@@ -0,0 +1,185 @@
...
+To find a link to a full list of database commands and corresponding
+parameters, see the :ref:`Additional Information section
+<php-addtl-info-runcommand>`.
+
+.. note:: Read Preference
+
+   The ``command()`` method does not
+   obey the read preference you might have set on your ``Database``
+   instance elsewhere in your code. By default, ``command()`` uses the ``primary``
+   read preference.
```
GromNaN, line 67: It seems that this first line could be longer (I hate manual line breaks).

source/write/run-command.txt
```diff
@@ -0,0 +1,185 @@
...
+         of the key used to sign the cluster time
+
+.. _php-command-example:
+
+Command Example
+---------------
+
+The following example uses the ``command()`` method to run
+the ``explain`` command for a ``count`` operation. The ``explain``
+command runs in the ``queryPlanner`` verbosity mode:
```
GromNaN, line 134: Note that the method `Collection::explain` exists.

source/write/run-command.txt
rustagir: added to api docs links

source/write/run-command.txt
rustagir: this probably arose from moving some sentences around!

source/write/run-command.txt
rustagir: I see. I can use a different example, such as `dbStats` or something similar?

rustagir:
Hi @GromNaN Please re-review PR when you have a moment!

source/includes/write/run-command.php
```diff
@@ -0,0 +1,17 @@
...
+$client = new MongoDB\Client($uri);
+
+// start-runcommand
+$database = $client->accounts;
+$command = ['dbStats' => 1];
+
+$result = $database->command($command);
+// end-runcommand
+
+var_dump(json_encode($result->toArray()));
```
jmikola, line 17: Consider moving this within the code snippet, since the RST also shows example output. As-is, the code above wouldn't output anything.

Also, consider using `echo json_encode(...);`, optionally adding `, PHP_EOL;`. I don't think other tutorial snippets used `var_dump()` when printing JSON-encoded documents as strings.

Also, [`dbStats`](https://www.mongodb.com/docs/manual/reference/command/dbStats/#output) returns its result in single document, so we can unpack that and include a comment to explain why we're doing so. Assuming we rename `$result` to `$cursor`.

```
// dbStats returns a single document, so print the first document in the cursor
echo json_encode($cursor->toArray()[0]), PHP_EOL;
```

In the RST, this would be a good reason to mention that some commands return multiple results in a cursor, while others return single result documents. The PHP driver handles this by always returning a Cursor from the `command()` helper in the interest of providing a consistent API that handles _both_ variants.

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+
+   The library provides wrapper methods for many database commands.
+   We recommend using these methods instead of executing database
+   commands when possible.
+   
+   To perform administrative tasks, use the :mongosh:`MongoDB Shell </>`
+   instead of the {+php-library+}. Calling the ``db.runCommand()``
+   method inside the shell is the preferred way to issue database
+   commands, as it provides a consistent interface between the shell and
+   drivers or libraries.
```
jmikola, line 38: I don't follow the last sentence here. How does `db.runCommand()` provide a "consistent interface between the shell and drivers or libraries"?

I would probably call out that the shell provides many more wrapper methods (aka helpers) for various administration commands: https://www.mongodb.com/docs/manual/reference/method/

And _those_ should probably also be preferred to running `db.runCommand()` or a driver's [command helper](https://github.com/mongodb/specifications/blob/master/source/run-command/run-command.md).

The selling point of the command helper is that it's a primitive tool to run any command for which the shell or driver doesn't already provide a wrapper/helper.

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+
+.. _php-execute-command:
+
+Execute a Command
+-----------------
+
+To run a database command, you must specify the command and any relevant
+parameters in a command document, then pass the command document to the
+``MongoDB\Database::command()`` method. This method returns a
+``Cursor`` object.
```
jmikola, line 48: This is also a good place to explain that commands that return a single document will still have that result packaged in a cursor via the driver API. And this ties into the `$cursor->toArray()[0]` code I suggested above.

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+
+The following code shows how you can use the ``command()``
+method on a ``MongoDB\Driver\Database`` instance to run the ``hello``
```
jmikola, line 51: This is a [MongoDB\Database](https://www.mongodb.com/docs/php-library/current/reference/class/MongoDBDatabase/) object. The PHP extension (i.e. `MongoDB\Driver` namespace) doesn't provide the common Client, Database, and Collection objects. I think you can use the `:phpclass:` directive to link to the API docs -- there should be prior art in other docs pages.

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+   You can set a read preference for command execution by setting one
+   in the options parameter, as shown in the following code:
+   
+   .. code-block:: php
+
+      $readPref = new MongoDB\Driver\ReadPreference('primaryPreferred');
+      $result = $database->command(
+          ['hello' => 1], 
+          ['readPreference' => $readPref]
+      );
```
jmikola, line 79: Any reason this code examples here and above are inline instead of included from the `run-command.php` file?

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+   have set on your ``Database`` instance elsewhere in your code. By
+   default, ``command()`` uses the ``primary`` read preference.
+   
+   You can set a read preference for command execution by setting one
+   in the options parameter, as shown in the following code:
+   
+   .. code-block:: php
+
+      $readPref = new MongoDB\Driver\ReadPreference('primaryPreferred');
+      $result = $database->command(
```
jmikola, line 76:
```suggestion
      $cursor = $database->command(
```

Makes the code self-documenting and reminds users they'll need to iterate it. You can also add `$cursor->toArray()[0]` below if you want to demonstrate printing the result. If so, consider a code comment above the `echo` line like I suggested earlier:

```
// hello returns a single document, so print the first document in the cursor
```

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+
+Response
+--------
+
+The ``command()`` method returns a ``Cursor`` object that contains
+the response from the database after the command has been executed.
+
+Each database command performs a different function, so the response
+content can vary depending on the command executed. However, every
+response contains a document with the following fields:
```
jmikola, line 94: For commands that return a cursor (e.g. `aggregate`), users are never going to have access to these fields because the driver converts the cursor envelope (e.g. ID and first batch of results) in the raw command response into an iterable cursor. This differs from commands that return a single document, which the driver would package as a cursor object containing a single result (which is why we use `$cursor->toArray()[0]`).

If you're going to talk about how "response content can vary", I would clearly explain this difference between commands and mention how the PHP driver handles this. So users are responsible for knowing the response format of their command (usually discussed in the server manual) so they can either iterate or just dump the first and only document in the cursor.

For an example of a cursor envelop in a command response, see the first output example in [listCollections](https://www.mongodb.com/docs/manual/reference/command/listCollections/#list-all-collections). Running `listCollections` through PHP's `command()` helper will never reveal that raw response, as we're going to construct a cursor from the first batch. There will also be no more results, since the cursor ID is zero (i.e. all results returned in the first batch). If the cursor ID was non-zero, then iterating beyond the first batch would utilize a `getMore` command to read more cursor data from the server.

I just noticed there's some confusion about this in the server manual and opened [DOCSP-43843](https://jira.mongodb.org/browse/DOCSP-43843) to report that.

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+     - Description
+
+   * - <command result>
+     - Fields specific to the database command. For example,
+       ``count`` returns the ``n`` field and ``explain`` returns the
+       ``queryPlanner`` field.
+
+   * - ``ok``
+     - Whether the command has succeeded (``1``)
+       or failed (``0``).
```
jmikola, line 110: You can note that the PHP driver will raise exceptions if the `ok` field is zero. It will likely be [CommandException](https://www.php.net/manual/en/class.mongodb-driver-exception-commandexception.php), which exposes a `getResultDocument()` method to obtain the command result document.

That behavior happens long before the PHP driver would try to unpack a command result with a cursor envelope, as a cursor envelope is part of a successful response and would already have an `ok: 1` field.

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+
+   * - ``$clusterTime``
+     - A document that contains the signed cluster time. Cluster time is a
+       logical time used for the ordering of operations.
+
+       This document contains the following fields:
+
+       - ``clusterTime``, the timestamp of the highest known cluster time for the member
+       - ``signature``, a document that contains the hash of the cluster time and the ID
+         of the key used to sign the cluster time
```
jmikola, line 126: Driver docs (including our specs) don't discuss the format of `$clusterTime`. It's only mentioned as being a BSON document. The type is intentionally opaque, as none of the driver APIs deal with these inner fields.

It looks like the content in this table was lifted from the [Go driver docs](https://www.mongodb.com/docs/drivers/go/current/fundamentals/run-command/#response).

I'll defer to you if it should be changed, but I'd probably suggest removing everything from "This document contains the following fields:" onward.

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+
+   * - ``ok``
+     - Whether the command has succeeded (``1``)
+       or failed (``0``).
+
+   * - ``operationTime``
+     - The logical time of the operation. MongoDB uses the
+       logical time to order operations. To learn more about this
+       concept, see our blog  post about the :website:`Global Logical
+       Clock </blog/post/transactions-background-part-4-the-global-logical-clock>`.
```
jmikola, line 116: `operationTime` is a BSON Timestamp type (not to be confused with a UTCDateTime). The PHP driver represents this as [MongoDB\BSON\Timestamp](https://www.php.net/manual/en/class.mongodb-bson-timestamp.php) if you want to use that.

I'm not sure about where this is discussed in the server manual, if at all, but relevant docs for these fields are in the following driver specs:

 * https://github.com/mongodb/specifications/blob/master/source/causal-consistency/causal-consistency.md#server-command-responses
 * https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.md

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+
+The output of this command includes information about the collections in
+the database and describes the amount and size of data stored across
+collections:
+
+.. code-block:: none
+
+   string(234) "[{"db":"accounts","collections":2,"views":0,"objects":5,"avgObjSize":22,
+   "dataSize":110,"storageSize":8192,"totalFreeStorageSize":0,"numExtents":0,"indexes":2,
+   "indexSize":8192,"indexFreeStorageSize":0,"fileSize":0,"nsSizeMB":0,"ok":1}]"
```
jmikola, line 151: Note: I suggested changing the output in the code file. You can avoid `var_dump()` and directly access the cursor's first and only document to make this more concise.

source/write/run-command.txt
```diff
@@ -0,0 +1,180 @@
...
+
+       This document contains the following fields:
+
+       - ``clusterTime``, the timestamp of the highest known cluster time for the member
+       - ``signature``, a document that contains the hash of the cluster time and the ID
+         of the key used to sign the cluster time
+
+.. _php-command-example:
+
+Command Example
```
jmikola, line 130: If you decide to improve this page to discuss commands that return a cursor, you can add another example for a cursor-bearing command like [listCollections](https://www.mongodb.com/docs/manual/reference/command/listCollections/). That would demonstrate how the driver automatically handles the cursor envelope and convey that the user never gets access to the original raw command result (it's technically still accessible via command monitoring if you dive into a CommandSucceededEvent).

source/write/run-command.txt
rustagir: thanks for pointing that out. Leaving the api links in the last section of the doc

source/write/run-command.txt
rustagir: I can move them over!

source/write/run-command.txt
rustagir: Changed the variable name - but in this simple example with no output, i think leaving out the iteration is ok.

source/write/run-command.txt
rustagir: I'll revise the introduction to that table, I will make the language more clear and indicate what the library behavior is here!

source/write/run-command.txt
jmikola: So long as the variable is `$cursor` and we have a previous example demonstrating `->toArray()[0]`, I'm OK with skipping it here. Users can logical infer how they should work with it.

source/write/run-command.txt
rustagir: Reworded the language here to make it clear that there's a hierarchy of preferred ways to perform tasks

source/write/run-command.txt
rustagir: I think this level of specificity is not necessary to include in this page

source/run-command.txt
```diff
@@ -0,0 +1,188 @@
...
+Response
+--------
+
+The ``command()`` method returns a ``Cursor`` object that contains
+the response from the database for the given command. Each database
+command performs a different function, so the response
+content can vary. Some command responses contain multiple result
+documents. In these situations, the library converts the cursor
+envelope in the raw command response, which includes the cursor ID and
+the first batch of results, into an iterable cursor.
```
jmikola, line 96: Consider adding something like the following to cover the general case for commands that _don't_ return cursors:

> For commands that return a single result document, that result will be made available as the first and only document in the cursor.

Since most cursor-returning commands have helpers, I'd consider this case to be more common.

source/run-command.txt
```diff
@@ -0,0 +1,188 @@
...
+.. _php-execute-command:
+
+Execute a Command
+-----------------
+
+To run a database command, you must specify the command and any relevant
+parameters in a command document, then pass the command document to the
+``MongoDB\Database::command()`` method. Many database commands return
+multiple result documents, so the ``command()`` method returns a
+``Cursor`` object that you can iterate through.
```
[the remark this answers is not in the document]
rustagir: its linked in the final section! but I updated the first use to be the full class name

source/run-command.txt
```diff
@@ -0,0 +1,188 @@
...
+Before you run a command, learn about the response format of the
+command so that your application either iterates through multiple
+results or extracts the first and only document in the cursor. See the
+:ref:`php-addtl-info-runcommand` section of this guide to find a link to
+the full list of database commands.
+
+If the number of documents in the command response is sufficiently large, you
+can run a :manual:`getMore </reference/command/getMore/>` command to
+retrieve the next batch of results from the cursor by using the cursor
+ID.
```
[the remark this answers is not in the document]
rustagir: I see - this applies only for the shell method I guess? Misunderstood a comment you made in the previous review.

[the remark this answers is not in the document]
jmikola: The behavior in PHPLIB's `Database::command()` (and PHPC's [`executeCommand()`](https://www.php.net/manual/en/mongodb-driver-manager.executecommand.php)) method is probably unique. Other drivers may have a separate method for the envelope handling (see: [runCursorCommand](https://github.com/mongodb/specifications/blob/master/source/run-command/run-command.md#runcursorcommand) in the driver spec).

AFAIK, [`db.runCommand()`](https://www.mongodb.com/docs/manual/reference/method/db.runCommand/) returns the raw response with no envelope unpacking, and I don't think the shell has a command cursor variant. I'm also not aware of any way in the shell to construct a cursor object from a raw response, so a script written for mongosh would probably need to manually execute `getMore` commands.

----

But more generally, if you have a cursor object in the driver or shell, you can trust that it will use `getMore` as needed to continue iteration. And `db.runCommand()` does not give you a cursor.

jmikola (approved):

PR #136, opened 2024-09-11
DOCSP-41963: Databases and collections

norareidy:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-41963
Staging - https://deploy-preview-136--docs-php-library.netlify.app/databases-collections/

## Self-Review Checklist

### DIFF
--- a/source/databases-collections.txt
+++ b/source/databases-collections.txt
@@ -45,6 +45,24 @@ The following example accesses a database named ``test_database``:
     :start-after: start-access-database
     :end-before: end-access-database
 
+Alternatively, you can implicitly call the ``MongoDB\Client::__get()`` magic method on
+a client object. This method allows you to select a database by using the syntax for
+accessing a class property. The following example uses this shorthand syntax to access
+the ``test_database`` database:
+
+.. literalinclude:: /includes/databases-collections/databases-collections.php
+    :language: php
+    :dedent:
+    :start-after: start-access-database-short
+    :end-before: end-access-database-short
+
+.. tip::
+
+   To learn more about ``__get()`` and PHP magic methods, see the following resources:
+
+   - :phpmethod:`MongoDB\Client::__get()` in the API documentation
+   - `Magic Methods <{+php-manual+}/language.oop5.magic.php>`__ in the PHP manual
+
 .. _php-db-coll-access-collection:
 
 Access a Collection
@@ -70,6 +88,20 @@ The following example accesses a collection named ``test_collection`` by using t
    If the provided collection name does not already exist in the database,
    MongoDB implicitly creates the collection when you first insert data
    into it.
+
+Alternatively, you can implicitly call the ``MongoDB\Database::__get()`` magic method on
+a database object. This method allows you to select a collection by using the syntax for
+accessing a class property. The following example uses this shorthand syntax to access
+the ``test_collection`` collection:
+
+.. literalinclude:: /includes/databases-collections/databases-collections.php
+    :language: php
+    :dedent:
+    :start-after: start-access-collection-short
+    :end-before: end-access-collection-short
+
+To learn more, see the :phpmethod:`MongoDB\Database::__get()`
+API documentation.
 
 .. _php-db-coll-create-collection:
 
@@ -100,7 +132,7 @@ You can query for a list of collections in a database by calling the
 cursor containing all collections in the database and their associated metadata.
 
 The following example calls the ``listCollections()`` method and iterates over
-the returned cursor to print the collections from the :ref:`php-db-coll-access-collection`
+the returned iterator to print the collections from the :ref:`php-db-coll-access-collection`
 and :ref:`php-db-coll-create-collection` examples:
 
 .. io-code-block::
@@ -128,9 +160,6 @@ and :ref:`php-db-coll-create-collection` examples:
          [type] => collection
          ...
       )
-
-For more information about iterating over a cursor, see the :ref:`php-cursors`
-guide. 
 
 Delete a Collection
 -------------------
@@ -164,22 +193,23 @@ read and write operations on a replica set by setting a **read concern** and a
 **write concern**.
 
 By default, databases inherit read and write settings from the ``MongoDB\Client``
-instance, and collections inherit them from the database. You can change
-these settings on your database by passing an options array to the
-``MongoDB\Client::selectDatabase()``. You can change these settings on your collection
-by passing an options array to the ``MongoDB\Client::selectCollection()`` method.
+instance. Collections inherit these settings from the ``MongoDB\Client`` or
+``MongoDB\Database`` instance on which the ``selectCollection()`` method is called.
+You can change these settings by passing an options array to the
+``MongoDB\Client::selectDatabase()``, ``MongoDB\Client::selectCollection()``, or
+``MongoDB\Database::selectCollection()`` methods.
 
 To change the read preference, read concern, and write concern of your database
 or collection, set the following options in the array parameter:
 
 - ``readPreference``: Sets the read preference. For a list of available read
-  preferences, see `MongoDB\\Driver\\ReadPreference <{+php-manual+}/class.mongodb-driver-readpreference>`__
+  preferences, see :php:`MongoDB\Driver\ReadPreference <mongodb-driver-readpreference>`
   in the extension API documentation.
 - ``readConcern``: Sets the read concern. For a list of available read
-  concerns, see `MongoDB\\Driver\\ReadConcern <{+php-manual+}/class.mongodb-driver-readconcern>`__
+  concerns, see :php:`MongoDB\Driver\ReadConcern <mongodb-driver-readconcern>`
   in the extension API documentation.
 - ``writeConcern``: Sets the write concern. For a list of available write
-  concerns, see `MongoDB\\Driver\\WriteConcern <{+php-manual+}/class.mongodb-driver-writeconcern>`__
+  concerns, see :php:`MongoDB\Driver\WriteConcern <mongodb-driver-writeconcern>`
   in the extension API documentation.
 
 .. tip::
@@ -217,8 +247,11 @@ array to ``selectCollection()``:
     :start-after: start-collection-settings
     :end-before: end-collection-settings
 
+Advanced Read Configurations 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 Tag Sets
-~~~~~~~~
+````````
 
 In {+mdb-server+}, you can apply key-value :manual:`tags
 </core/read-preference-tags/>` to replica-set
@@ -243,7 +276,7 @@ center (``'dc' => 'sf'``):
     :end-before: end-tag-set
 
 Local Threshold
-~~~~~~~~~~~~~~~
+```````````````
 
 If multiple replica-set members match the read preference and tag sets you specify,
 the {+php-library+} reads from the nearest replica-set members, chosen according to
--- a/source/databases-collections.txt
+++ b/source/databases-collections.txt
@@ -249,7 +249,7 @@ If multiple replica-set members match the read preference and tag sets you speci
 the {+php-library+} reads from the nearest replica-set members, chosen according to
 their ping time.
 
-By default, the library only uses members whose ping times are within 15 milliseconds
+By default, the library uses only members whose ping times are within 15 milliseconds
 of the nearest member for queries. To distribute reads between members with
 higher latencies, pass an options array to the ``MongoDB\Client`` constructor that
 sets the ``localThresholdMS`` option.
--- a/source/databases-collections.txt
+++ b/source/databases-collections.txt
@@ -264,10 +264,13 @@ pass them as a parameter to your ``MongoDB\Driver\ReadPreference`` class
 constructor. Then, set the ``MongoDB\Driver\ReadPreference`` object as
 the value of the ``readPreference`` database option.
 
-The following code example sets the ``readPreference`` option to a tag set
-that instructs ``test_database`` to prefer reads from the New York data
-center (``'dc' => 'ny'``) and to fall back to the San Francisco data
-center (``'dc' => 'sf'``):
+This code example sets the ``readPreference`` option to a tag set
+that instructs ``test_database`` to prefer reads from secondary replica set
+members in the following order:
+
+1. Members from the New York data center (``['dc' => 'ny']``)
+#. Members from the San Francisco data center (``['dc' => 'sf']``)
+#. Any secondary members (``[]``)
 
 .. literalinclude:: /includes/databases-collections/databases-collections.php
     :language: php
--- a/source/databases-collections.txt
+++ b/source/databases-collections.txt
@@ -1,0 +1,276 @@
+.. _php-databases-collections:
+
+=========================
+Databases and Collections
+=========================
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. facet::
+   :name: genre
+   :values: reference
+
+.. meta::
+   :keywords: table, row, organize, storage, code example
+
+Overview
+--------
+
+In this guide, you can learn how to use MongoDB databases and
+collections with the {+php-library+}.
+
+MongoDB organizes data into a hierarchy of the following levels:
+
+- **Databases**: Top-level data structures in a MongoDB deployment that store collections.
+- **Collections**: Groups of MongoDB documents. They are analogous to tables in relational databases.
+- **Documents**: Units that store literal data such as string, numbers, dates, and other embedded documents.
+  For more information about document field types and structure, see the
+  :manual:`Documents </core/document/>` guide in the {+mdb-server+} manual.
+
+Access a Database
+-----------------
+
+Access a database by passing the database name to the ``MongoDB\Client::selectDatabase()``
+method.
+
+The following example accesses a database named ``'test_database'``:
+
+.. literalinclude:: /includes/databases-collections/databases-collections.php
+    :language: php
+    :dedent:
+    :start-after: start-access-database
+    :end-before: end-access-database
+
+.. _php-db-coll-access-collection:
+
+Access a Collection
+-------------------
+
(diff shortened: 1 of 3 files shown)


source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+    :dedent:
+    :start-after: start-drop-collection
+    :end-before: end-drop-collection
+
+.. warning:: Dropping a Collection Deletes All Data in the Collection
+
+   Dropping a collection from your database permanently deletes all
+   documents and all indexes within that collection.
+
+   Drop a collection only if the data in it is no longer needed.
```
mongoKart, line 154: S:

```suggestion
   Drop a collection only if you no longer need the data in it.
```

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+The following example creates a collection named ``example_collection``:
+
+.. literalinclude:: /includes/databases-collections/databases-collections.php
+    :language: php
+    :dedent:
+    :start-after: start-create-collection
+    :end-before: end-create-collection
+
+You can specify collection options, such as maximum size and document
+validation rules, by passing them as an array parameter to ``createCollection()``.
```
mongoKart, line 91: S:

```suggestion
validation rules, by passing them as an array to the ``createCollection()`` method.
```

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+Configure Read and Write Operations
+-----------------------------------
+
+You can control how the library routes read operations by setting a **read preference**.
+You can also control options for how the library waits for acknowledgment of
+read and write operations on a replica set by setting a **read concern** and a
+**write concern**.
+
+By default, databases inherit read and write settings from the ``MongoDB\Client``
+instance, and collections inherit them from the database. However, you can change
```
mongoKart, line 167:
```suggestion
instance, and collections inherit them from the database. You can change
```

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+or collection, set the following options in the array parameter:
+
+- ``readPreference``: Sets the read preference. For a list of available read
+  preferences, see `MongoDB\\Driver\\ReadPreference <{+php-manual+}/class.mongodb-driver-readpreference>`__
+  in the extension API documentation.
+- ``readConcern``: Sets the read concern. For a list of available read
+  concerns, see `MongoDB\\Driver\\ReadConcern <{+php-manual+}/class.mongodb-driver-readconcern>`__
+  in the extension API documentation.
+- ``writeConcern``: Sets the write concern. For a list of available write
+  concerns, see `MongoDB\\Driver\\ReadConcern <{+php-manual+}/class.mongodb-driver-writeconcern>`__
```
mongoKart, line 182:
```suggestion
  concerns, see `MongoDB\\Driver\\WriteConcern <{+php-manual+}/class.mongodb-driver-writeconcern>`__
```

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+  in the extension API documentation.
+
+.. tip::
+
+   To learn more about the read and write settings, see the following guides in the
```
mongoKart, line 187:
```suggestion
   To learn more about read preferences and read and write concerns, see the following guides in the
```

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+
+   - :manual:`Read Preference </core/read-preference/>`
+   - :manual:`Read Concern </reference/read-concern/>`
+   - :manual:`Write Concern </reference/write-concern/>`
+
+Database Configuration Example
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following example shows how to set the read preference, read concern, and
+write preference of a database called ``test_database`` by passing an options
```
mongoKart, line 198:
```suggestion
write concern of a database called ``test_database`` by passing an options
```

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+    :language: php
+    :dedent:
+    :start-after: start-database-settings
+    :end-before: end-database-settings
+
+Collection Configuration Example
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following example shows how to set the read preference, read concern, and
+write preference of a collection called ``test_collection`` by passing an options
```
mongoKart, line 211:
```suggestion
write concern of a collection called ``test_collection`` by passing an options
```

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+~~~~~~~~~~~~~~~
+
+If multiple replica-set members match the read preference and tag sets you specify,
+the {+php-library+} reads from the nearest replica-set members, chosen according to
+their ping time.
+
+By default, the library uses only those members whose ping times are within 15 milliseconds
+of the nearest member for queries. To distribute reads between members with
+higher latencies, pass an options array to the ``MongoDB\Client`` constructor that
+sets the ``localThresholdMS`` option.
```
mongoKart, line 255: S: this might be slightly clearer

```suggestion
The library uses only those members whose ping times are within the local
threshold of the nearest member for queries. By default,
the local threshold is 15 milliseconds of latency. To distribute reads among members with
higher latencies, pass an options array to the ``MongoDB\Client`` constructor that
sets the ``localThresholdMS`` option.
```

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+
+The following example specifies a local threshold of 35 milliseconds:
+
+.. literalinclude:: /includes/databases-collections/databases-collections.php
+    :language: php
+    :dedent:
+    :start-after: start-local-threshold
+    :end-before: end-local-threshold
+
+In the preceding example, the {+php-library+} distributes reads between matching members
```
mongoKart, line 265: https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#a:~:text=00%20p.m.-,among,-between%20When%20referring

```suggestion
In the preceding example, the {+php-library+} distributes reads among matching members
```

mongoKart (changes requested):
a few small issues and suggestions. page looks really good, though!

mongoKart (changes requested):
lgtm w/ 1 fix

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+.. literalinclude:: /includes/databases-collections/databases-collections.php
+    :language: php
+    :dedent:
+    :start-after: start-create-collection
+    :end-before: end-create-collection
+
+You can specify collection options, such as maximum size and document
+validation rules, by passing them as an array to the ``createCollection()`` method.
+For a full list of optional parameters, see the `API documentation
+<{+api+}/method/MongoDBDatabase-createCollection/#parameters>`__.
```
jmikola, line 93: I assume `:phpmethod: is not used here because you want to deep link to the parameters heading?

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+
+Get a List of Collections
+-------------------------
+
+You can query for a list of collections in a database by calling the
+``MongoDB\Database::listCollections()`` method. The method returns a
+cursor containing all collections in the database and their associated metadata.
+
+The following example calls the ``listCollections()`` method and iterates over
+the returned cursor to print the collections from the :ref:`php-db-coll-access-collection`
```
jmikola, line 103:
```suggestion
the returned iterator to print the collections from the :ref:`php-db-coll-access-collection`
```

Although the [`listCollections` command](https://www.mongodb.com/docs/manual/reference/command/listCollections/) returns a cursor, PHPLIB abstracts that with another iterator. This allows us to do things like [cache the results](https://github.com/mongodb/mongo-php-library/blob/v1.x/src/Model/CachingIterator.php) and return typed model classes (e.g. CollectionInfo).

Caching is helpful if users want to iterate the result multiple times, as the original cursor can only be iterated once as it streams results from the server.

Also note that the special iterator classes returned by the database, collection, and index enumeration methods are being deprecated and will be removed in 2.0 (see: [PHPLIB-1522](https://jira.mongodb.org/browse/PHPLIB-1522)). The individual result classes (e.g. CollectionInfo) will remain, though. In order to future-proof the docs, I propose we just talk about these methods returning an iterator.

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+      )
+      MongoDB\Model\CollectionInfo Object
+      (
+         [name] => test_collection
+         [type] => collection
+         ...
+      )
+
+For more information about iterating over a cursor, see the :ref:`php-cursors`
+guide.
```
jmikola, line 133:
```suggestion
```

This can be removed since none of these enumeration methods return a cursor.

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+You can control how the library routes read operations by setting a **read preference**.
+You can also control options for how the library waits for acknowledgment of
+read and write operations on a replica set by setting a **read concern** and a
+**write concern**.
+
+By default, databases inherit read and write settings from the ``MongoDB\Client``
+instance, and collections inherit them from the database. You can change
+these settings on your database by passing an options array to the
+``MongoDB\Client::selectDatabase()``. You can change these settings on your collection
+by passing an options array to the ``MongoDB\Client::selectCollection()`` method.
```
jmikola, line 170: Regarding collection inheritance, note that if users call `Client::selectCollection()` they would effectively inherit the Client's options since an interstitial Database object is never created.

`Database::selectCollection()` isn't even mentioned in this paragraph, although it was in a previous section. I'll note that the code example only demonstrates `Database::selectCollection()`, so perhaps you want to focus on that here instead of the Client method.

I'll defer to you if any of this requires some attention.

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+
+To change the read preference, read concern, and write concern of your database
+or collection, set the following options in the array parameter:
+
+- ``readPreference``: Sets the read preference. For a list of available read
+  preferences, see `MongoDB\\Driver\\ReadPreference <{+php-manual+}/class.mongodb-driver-readpreference>`__
+  in the extension API documentation.
+- ``readConcern``: Sets the read concern. For a list of available read
+  concerns, see `MongoDB\\Driver\\ReadConcern <{+php-manual+}/class.mongodb-driver-readconcern>`__
```
jmikola, line 179: Existing docs use the `:php:` directive. For example:

```
:php:`MongoDB\Driver\ReadConcern <class.mongodb-driver-readconcern>`
```

Are there plans to eliminate that? If not, should you use it for PHP.net links?

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+By default, the {+php-library+} ignores tags when choosing a member
+to read from. To instruct the {+php-library+} to prefer certain tags,
+pass them as a parameter to your ``MongoDB\Driver\ReadPreference`` class
+constructor. Then, set the ``MongoDB\Driver\ReadPreference`` object as
+the value of the ``readPreference`` database option.
+
+The following code example sets the ``readPreference`` option to a tag set
+that instructs ``test_database`` to prefer reads from the New York data
+center (``'dc' => 'ny'``) and to fall back to the San Francisco data
+center (``'dc' => 'sf'``):
```
jmikola, line 237: It's been a while since I revisited tag set docs, but I recall examples typically included an empty tag set as a fallback.

As-is, server selection will fail if there are no servers in either `dc:ny` or `dc:sf`. Just want to confirm that is intentional.

source/databases-collections.txt
```diff
@@ -0,0 +1,286 @@
...
+write concern of a collection called ``test_collection`` by passing an options
+array to ``selectCollection()``:
+
+.. literalinclude:: /includes/databases-collections/databases-collections.php
+    :language: php
+    :dedent:
+    :start-after: start-collection-settings
+    :end-before: end-collection-settings
+
+Tag Sets
```
jmikola, line 220: Should "Tag Sets" and "Local Threshold" be grouped under a "Read Preference" heading?

source/includes/databases-collections/databases-collections.php
```diff
@@ -0,0 +1,93 @@
+<?php
+require 'vendor/autoload.php';
+
+use MongoDB\BSON\Document;
```
jmikola, line 4:
```suggestion
use MongoDB\BSON\Document;
```

This class isn't referenced in the code.

source/includes/databases-collections/databases-collections.php
```diff
@@ -0,0 +1,93 @@
...
+$client = new MongoDB\Client($uri);
+
+// Accesses the "test_database" database
+// start-access-database
+$db = $client->selectDatabase('test_database');
+// end-access-database
+
+// Accesses the "test_collection" collection
+// start-access-collection
+$collection = $client->test_database->selectCollection('test_collection');
```
jmikola, line 19: Since this file is using a mix of the shorthand syntax (e.g. `$client->test_database`) and explicit methods, would it be helpful to include notes in the RST that talks about how PHPLIB implements the [`__get() magic method`](https://www.php.net/manual/en/language.oop5.overloading.php#language.oop5.overloading.members) on Client and Database methods as an alternative way to access a child database or collection, respectively?

While it's implied from the examples on this page, you could also explicitly call out that the property access shorthand doesn't provide any way to specify options (in contrast to the explicit select methods).

source/includes/databases-collections/databases-collections.php
```diff
@@ -0,0 +1,93 @@
...
+
+// Explicitly creates the "example_collection" collection
+// start-create-collection
+$result = $client->test_database->createCollection('example_collection');
+// end-create-collection
+
+// Lists the collections in the "test_database" database
+// start-find-collections
+foreach ($client->test_database->listCollections() as $collectionInfo) {
+    print_r($collectionInfo) . PHP_EOL;
```
jmikola, line 30: I assume `print_r()` might be preferred here since `var_dump()` output is a bit messier.

source/includes/databases-collections/databases-collections.php
```diff
@@ -0,0 +1,93 @@
...
+}
+// end-find-collections
+
+// Deletes the "test_collection" collection
+// start-drop-collection
+$client->test_database->dropCollection('test_collection');
+// end-drop-collection
+
+// Sets read and write settings for the "test_database" database
+// start-database-settings
```
jmikola, line 40: I noticed the RST doesn't talk about specifying any of these options on the client by way of the connection string. Is that to be covered by some other tutorial?

source/includes/databases-collections/databases-collections.php
```diff
@@ -0,0 +1,93 @@
...
+
+$collection = $client->selectCollection('test_database', 'test_collection', [
+    'readPreference' => $readPreference,
+    'readConcern' => $readConcern,
+    'writeConcern' => $writeConcern,
+]);
+
+// end-collection-settings
+
+// Instructs the library to prefer New York data center reads using a tag set
```
jmikola, line 66: In a previous PR you mentioned that these comments are mainly used for influence AI analysis of the code example that follows. If so, should this description be more accurate?

You're actually selecting a secondary server in either NY or SF. The detail about it being a secondary and falling back to SF are what's missing.

source/includes/databases-collections/databases-collections.php
```diff
@@ -0,0 +1,93 @@
...
+
+// end-tag-set
+
+// Instructs the library to distribute reads between members within 35 milliseconds
+// of the closest member's ping time
+// start-local-threshold
+$options = [
+    'replicaSet' => 'repl0',
+    'readPreference' => new ReadPreference(ReadPreference::RP_SECONDARY_PREFERRED),
+    'localThresholdMS' => 35
```
jmikola, line 89:
```suggestion
    'localThresholdMS' => 35,
```

source/includes/databases-collections/databases-collections.php
```diff
@@ -0,0 +1,93 @@
+<?php
+require 'vendor/autoload.php';
+
+use MongoDB\BSON\Document;
+use MongoDB\Driver\ReadConcern;
+use MongoDB\Driver\ReadPreference;
+use MongoDB\Driver\WriteConcern;
+
+$uri = getenv('MONGODB_URI') ?: throw new RuntimeException('Set the MONGODB_URI variable to your Atlas URI that connects to the sample dataset');
```
jmikola, line 9: "that connects to the sample dataset" doesn't seem relevant for this file.

source/databases-collections.txt
norareidy: Good to know, I wasn't aware of that directive - updated

source/databases-collections.txt
norareidy: yes!

source/databases-collections.txt
norareidy: Added an empty tag set

source/includes/databases-collections/databases-collections.php
norareidy: Yeah users don't see this, but I reworded slightly

source/databases-collections.txt
jmikola: Noted that "Advanced Read Configurations" was added.

source/includes/databases-collections/databases-collections.php
```diff
@@ -63,13 +72,15 @@
 
 // end-collection-settings
 
-// Instructs the library to prefer New York data center reads using a tag set
+// Instructs the library to prefer reads from secondary replica set members
+// located in New York but fall back to those in San Francisco
```
jmikola, line 76:
```suggestion
// located in New York, followed by a secondary in San Francisco, and
// lastly fall back to any secondary.
```

This captures the effect of the empty tag set.

Please adjust the RST paragraph above this code example accordingly. If you want to borrow some language from the existing server manual, there's a nicely worded tip in [Order of Tag Matching](https://www.mongodb.com/docs/v5.2/core/read-preference-tags/#order-of-tag-matching).

jmikola (commented):
One last correction regarding the empty tag set.

jmikola (approved):

PR #137, opened 2024-09-11
DOCSP-41995: transaction

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-41995>
Staging - https://deploy-preview-137--docs-php-library.netlify.app/write/transaction/

## Self-Review Checklist


source/includes/write/transaction.php
```diff
@@ -0,0 +1,53 @@
...
+    );
+
+    $receipts->insertOne([
+        'account_id' => $accountId,
+        'description' => $summary,
+        'timestamp' => new MongoDB\BSON\UTCDateTime(),
+    ],
+    ['session' => $session]);
+
+    echo 'Successfully performed transaction!' , "\n";
```
norareidy, line 41: S: I've been using PHP_EOL instead of \n to avoid double quotes
```suggestion
    echo 'Successfully performed transaction!' , PHP_EOL;
```

source/write/transaction.txt
```diff
@@ -0,0 +1,168 @@
...
+
+Then, run the following code to perform the transaction. This code
+completes the following actions:
+
+1. Creates a session from the client by using the ``startSession()`` method.
+#. Calls the ``with_transaction()`` method to manage the transaction,
+   passing the session and the callback as parameters.
+
+.. io-code-block::
+
```
norareidy, line 116: S: make this code example copyable
```suggestion
.. io-code-block::
   :copyable:
```

source/write/transaction.txt
```diff
@@ -0,0 +1,168 @@
...
+Additional Information
+----------------------
+
+To learn more about the concepts mentioned in this guide, see the
+following pages in the {+mdb-server+} manual:
+
+- :manual:`Transactions </core/transactions/>`
+- :manual:`Server Sessions </reference/server-sessions/>`
+- :manual:`Read Isolation, Consistency, and Recency
+  </core/read-isolation-consistency-recency/#causal-consistency>`
```
norareidy, line 138: Q: did you intentionally link to the Casual Consistency section of this page? I think you can just link to the main page
```suggestion
- :manual:`Read Isolation, Consistency, and Recency
  </core/read-isolation-consistency-recency/>`
```

source/write/transaction.txt
```diff
@@ -0,0 +1,168 @@
...
+- :phpclass:`MongoDB\Client`
+- :phpmethod:`MongoDB\Client::startSession()`
+- :phpmethod:`MongoDB\Collection::updateOne()`
+- :phpmethod:`MongoDB\Collection::insertOne()`
+
+To learn more about the ``Session`` class and methods,
+see the following {+extension-short+} API documentation:
+
+- `MongoDB\\Driver\\Session
+  <https://www.php.net/manual/en/class.mongodb-driver-session.php>`__
```
norareidy, line 162: S: you can use the {+php-manual+} constant in these links
```suggestion
  <{+php-manual+}/class.mongodb-driver-session.php>`__
```

norareidy: applies to all

source/write/transaction.txt
```diff
@@ -0,0 +1,168 @@
...
+         transaction was committed or ended.
+
+.. _php-txn-example:
+
+Transaction Example
+-------------------
+
+The following code defines the ``transferTxn()`` callback function that
+modifies data in the ``saving_accounts``, ``checking_accounts``, and ``receipts``
+collections of the ``bank`` database:
```
norareidy, line 100: S: not a necessary change, but there's a lot going on in this code example and it might be helpful to give more detail about how you're modifying those collections in a bulleted list

source/write/transaction.txt
```diff
@@ -0,0 +1,168 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to use the {+driver-short+} to perform
+**transactions**. Transactions allow you to perform a series of operations
+that change data only if the entire transaction is committed.
+If any operation in the transaction does not succeed, the driver stops the
```
norareidy, line 26: S: I think "library" is a better word than "driver" for PHP
```suggestion
If any operation in the transaction does not succeed, the library stops the
```

norareidy (approved):
LGTM + some suggestions

source/write/transaction.txt
rustagir: good point!

rustagir:
Hi @GromNaN Please take a look at this PR when you have a moment!

source/includes/write/transaction.php
```diff
@@ -0,0 +1,54 @@
...
+    $summary = sprintf(
+        "SAVINGS +%u CHECKING -%u", $transferAmount, $transferAmount
+    );
+
+    $receipts->insertOne([
+        'account_id' => $accountId,
+        'summary' => $summary,
+        'timestamp' => new MongoDB\BSON\UTCDateTime(),
+    ],
+    ['session' => $session]);
```
jmikola, line 39:
```suggestion
    $receipts->insertOne(
        [
            'account_id' => $accountId,
            'summary' => $summary,
            'timestamp' => new MongoDB\BSON\UTCDateTime(),
        ],
        ['session' => $session],
    );
```

Optional code formatting suggestion to avoid having anything _within_ the method aligned to the parent level.

source/includes/write/transaction.php
```diff
@@ -0,0 +1,54 @@
...
+    
+    $saving->updateOne(
+        ['account_id' => $accountId],
+        ['$inc' => ['balance' => $transferAmount]],
+        ['session' => $session]
+    );
+
+    $summary = sprintf(
+        "SAVINGS +%u CHECKING -%u", $transferAmount, $transferAmount
+    );
```
jmikola, line 32:
```suggestion
    $summary = sprintf('SAVINGS +%1$u CHECKING -%1$u', $transferAmount);
```

`printf()`-style functions support an argument number to reduce repetition. See: https://www.php.net/manual/en/function.sprintf.php

source/includes/write/transaction.php
```diff
@@ -0,0 +1,54 @@
...
+
+    $receipts->insertOne([
+        'account_id' => $accountId,
+        'summary' => $summary,
+        'timestamp' => new MongoDB\BSON\UTCDateTime(),
+    ],
+    ['session' => $session]);
+
+    echo 'Successfully performed transaction!' , PHP_EOL;
```
jmikola, line 41:
```suggestion
    echo 'Successfully performed transaction!', PHP_EOL;
```

source/includes/write/transaction.php
```diff
@@ -0,0 +1,54 @@
...
+    echo 'Summary: ', $summary, PHP_EOL;
+};
+// end-callback
+
+// begin-txn
+$session = $client->startSession();
+
+try {
+    MongoDB\with_transaction($session, $callback);
+} catch (Exception $e) {
```
jmikola, line 51:
```suggestion
} catch (MongoDB\Driver\Exception\RuntimeException $e) {
```

Optional suggestion, but the best practice would be to limit the catch to what can be expected at runtime. The general Exception class would cover _anything_ from PHP, including something like an error passing an invalid parameter or option in the driver -- which is better left to blow up in the user's face to alert them to a coding mistake.

source/write/transaction.txt
```diff
@@ -0,0 +1,183 @@
...
+atomicity, consistency, isolation, and durability. MongoDB guarantees
+that the data involved in your transaction operations remains
+consistent, even if the operations encounter unexpected errors.
+
+When using the {+php-library+}, you can create a new session from a
+``MongoDB\Client`` instance. Then, you can use the resulting
+``MongoDB\Driver\Session`` instance to perform transactions. You can
+improve your app's performance by reusing your client for multiple
+sessions and transactions instead of instantiating a new client each
+time.
```
jmikola, line 44: It's very uncommon that a PHP app would would create multiple clients. Also, it's quite possible that multiple Clients could share the same underlying libmongoc client anyway if they're constructed with the same arguments (see: [Connection handling and persistence](https://www.php.net/manual/en/mongodb.connection-handling.php)).

I don't think we need this sentence.

source/write/transaction.txt
```diff
@@ -0,0 +1,183 @@
...
+   ``Client`` that created it. Using a ``Session`` with a
+   different ``Client`` results in operation errors.
+
+Methods
+-------
+
+Create a ``Session`` by using the ``MongoDB\Client::startSession()``
+method on your ``Client`` instance. The {+php-library+} provides a
+Convenient Transaction API to manage the transaction lifecyle. Use the
+``MongoDB\with_transaction()`` method to run custom callback within a
```
jmikola, line 58:
```suggestion
``MongoDB\with_transaction()`` function to run custom callback within a
```

Technical nit: method is a function on a class. This is a standalone utility function not tied to any class.

source/write/transaction.txt
```diff
@@ -0,0 +1,183 @@
...
+   Use a ``Session`` only in operations running on the
+   ``Client`` that created it. Using a ``Session`` with a
+   different ``Client`` results in operation errors.
+
+Methods
+-------
+
+Create a ``Session`` by using the ``MongoDB\Client::startSession()``
+method on your ``Client`` instance. The {+php-library+} provides a
+Convenient Transaction API to manage the transaction lifecyle. Use the
```
jmikola, line 57: I noticed the name is emphasized in the Node docs: https://www.mongodb.com/docs/drivers/node/current/usage-examples/transaction-conv/

Should that be done here?

source/write/transaction.txt
```diff
@@ -0,0 +1,183 @@
...
+``MongoDB\with_transaction()`` method to run custom callback within a
+transaction. The ``with_transaction()`` method starts the transaction,
+then either commits it or ends it if there are errors. The
```
jmikola, line 60:
```suggestion
then either commits it or ends it if there are errors. The
```

This sentence overlooks the actual benefit of the Convenient Transaction API. See: https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.md#abstract

We should mention that it's capable of automatically retrying the transaction for certain errors such as a transient write conflict.

source/write/transaction.txt
```diff
@@ -0,0 +1,183 @@
...
+     - | Starts a new transaction on this session. The session
+         must be passed into each operation within the transaction, or
+         the operation will run outside of the transaction.
+       |
+       | You can set transaction options by passing an options parameter.
+
+   * - ``commitTransaction()``
+     - | Commits the active transaction for this session. This method returns an
+         error if there is no active transaction for the session or the
+         transaction was previously ended.
```
jmikola, line 86: Notably, a commit can fail due to other transient errors, such as a write conflict. That would be more common as a runtime error, vs. a user calling this when there is no active transaction (likely a poorly written application, and a mistake they'd make once and correct).

source/write/transaction.txt
```diff
@@ -0,0 +1,183 @@
...
+
+This example defines a callback function that
+modifies data in the collections of the ``bank`` database for a
+banking transaction. The code performs the following actions:
+
+- Creates ``Collection`` instances to access the target
+  collections.
+- Specifies the account number and amount to be transferred between
+  accounts.
+- Defines the callback function, passing the ``Session`` instance as a
```
jmikola, line 106:
```suggestion
- Defines the callback function, which receives a ``Session`` instance as a
```

Technically, the `with_transaction()` function passes the parameter each time it invokes the callback. This is the same Session that the application passes into `with_transaction()`.

source/write/transaction.txt
```diff
@@ -0,0 +1,183 @@
...
+
+API Documentation
+~~~~~~~~~~~~~~~~~
+
+To learn more about the methods and types mentioned in this
+guide, see the following API documentation:
+
+- :phpclass:`MongoDB\Client`
+- :phpmethod:`MongoDB\Client::startSession()`
+- :phpmethod:`MongoDB\with_transaction()`
```
jmikola, line 169: I'll note that the docs use a common `:phpmethod:` directive here, even though this is technically a function. Nothing to address here -- just wanted to point that out.

source/includes/write/transaction.php
rustagir: used a code formatter to fix these issues

source/includes/write/transaction.php
rustagir: excellent!

source/write/transaction.txt
rustagir: replaced in all instances

source/write/transaction.txt
rustagir: revised this section

source/write/transaction.txt
rustagir: added that clause

source/write/transaction.txt
```diff
@@ -0,0 +1,201 @@
...
+         must be passed into each operation within the transaction, or
+         the operation will run outside of the transaction.
+       |
+       | You can set transaction options by passing an options parameter.
+
+   * - ``commitTransaction()``
+     - | Commits the active transaction for this session. This method returns an
+         error if there is no active transaction for the session, the
+         transaction was previously ended, or if there is a write
+         conflict.
```
jmikola, line 108:
```suggestion
         transaction was previously ended, or if there is a write conflict.
```

If you want to avoid extra wrapping.

jmikola (approved):
Optional word wrap suggestion. Content LGTM.

PR #139, opened 2024-09-12
DOCSP-41990: Authentication mechanisms

norareidy:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-41990
Staging - https://deploy-preview-139--docs-php-library.netlify.app/security/authentication/

## Self-Review Checklist

- [ ] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?


source/security/authentication.txt
```diff
@@ -0,0 +1,403 @@
...
+  {+php-library+} authenticates against the database in the connection
+  URI, if you include one. If you don't, it authenticates against the ``admin`` database.
+- ``authMechanism``: Set to ``'SCRAM-SHA-256'``.
+
+You can set these options in two ways: by passing an options array to the
+``MongoDB\Client`` constructor or through parameters in your connection string.
+
+.. tabs::
+
+   .. tab:: MongoDB\Client
```
mongoKart, line 61: Showing up with only one slash. Applies throughout. I think this should work:

```suggestion
   .. tab:: MongoDB\\Client
```

source/security/authentication.txt
```diff
@@ -0,0 +1,403 @@
...
+         $client = new Client($uri);
+
+.. _php-mongodb-aws:
+
+MONGODB-AWS
+-----------
+
+.. important::
+
+   The MONGODB-AWS authentication mechanism requires MongoDB v4.4 or later.
```
mongoKart, line 187:
```suggestion
   The MONGODB-AWS authentication mechanism requires {+mdb-server+} v4.4 or later.
```

source/security/authentication.txt
```diff
@@ -0,0 +1,403 @@
...
+
+The MONGODB-AWS authentication mechanism uses AWS IAM (Amazon Web Services Identity and
+Access Management) or AWS Lambda credentials to authenticate your application. To
+authenticate using this mechanism, first create a user with an associated Amazon Resource
```
mongoKart, line 191:
```suggestion
Access Management) or AWS Lambda credentials to authenticate your application. To
use this mechanism to authenticate, first create a user with an associated Amazon Resource
```

source/security/authentication.txt
```diff
@@ -0,0 +1,403 @@
...
+Name (ARN) on the ``$external`` database, then specify ``'MONGODB-AWS'`` as the value of
+the ``authMechanism`` connection option.
```
mongoKart, line 193: S:

```suggestion
Name (ARN) on the ``$external`` database. Then, specify ``'MONGODB-AWS'`` as the value of
the ``authMechanism`` connection option.
```

source/security/authentication.txt
```diff
@@ -0,0 +1,403 @@
...
+#. AWS EKS ``AssumeRoleWithWebIdentity`` request
+#. ECS container metadata
+#. EC2 instance metadata
+
+The following sections describe how to use the {+php-library+} to retrieve credentials from
+these sources and use them to authenticate your application.
+
+.. _php-mongodb-aws-credentials:
+
+``MongoDB\Client`` Credentials
```
mongoKart, line 210: The Style Guide technically prohibits using monospace in titles and headings (though I think this is a bad rule)

```suggestion
MongoDB\Client Credentials
```

source/security/authentication.txt
```diff
@@ -0,0 +1,403 @@
...
+  it in a connection URI.
+- ``password``: The password to authenticate. Percent-encode this value before including
+  it in a connection URI.
+- ``authSource``: The MongoDB database to authenticate against. By default, the
+  {+php-library+} authenticates against the database in the connection
+  URI, if you include one. If you don't, it authenticates against the ``admin`` database.
+- ``authMechanism``: Set to ``'SCRAM-SHA-256'``.
+
+You can set these options in two ways: by passing an options array to the
+``MongoDB\Client`` constructor or through parameters in your connection string.
```
mongoKart, line 57: S: Use "connection URI" to be consistent with rest of page. Applies throughout

```suggestion
``MongoDB\Client`` constructor or through parameters in your connection URI.
```

source/security/authentication.txt
```diff
@@ -0,0 +1,403 @@
...
+   
+   AWS Lambda runtimes can automatically set these environment variables during
+   initialization. For more information about using environment variables in an AWS Lambda
+   environment, see `Using Lambda environment variables
+   <https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime>`__
+   in the AWS documentation.
+
+.. _php-mongodb-aws-oidc:
+
+``AssumeRoleWithWebIdentity`` Request
```
mongoKart, line 322:
```suggestion
AssumeRoleWithWebIdentity Request
```

source/security/authentication.txt
```diff
@@ -0,0 +1,403 @@
...
+To view an example that sets the ``authMechanism`` option, see the :ref:`authMechanism example
+<php-mongodb-aws-env-example>` on this page.
+
+.. note::
+
+   If you set any of the environment variables from the preceding AWS authentication
+   methods, the {+php-library+} attempts to retrieve credentials by using those
+   methods before attempting to retrieve them from an EC2 instance. To attempt
+   to retrieve credentials only from an EC2 instance, ensure that the
+   environment variables are not set.
```
mongoKart, line 386: S: confusing use of "method"

```suggestion
   The {+php-library+} retrieves credentials from an EC2 instance only if the
   environment variables described earlier in this section are not set.
```

Consider linking to that earlier section, too

mongoKart (changes requested):
Just some small things!

source/security/authentication.txt
norareidy: agreed (but fixed)

mongoKart (approved):
LGTMeeeeeee

source/security/authentication.txt
```diff
@@ -0,0 +1,401 @@
...
+-----------
+
+.. important::
+
+   The MONGODB-AWS authentication mechanism requires {+mdb-server+} v4.4 or later.
+
+The MONGODB-AWS authentication mechanism uses AWS IAM (Amazon Web Services Identity and
+Access Management) or AWS Lambda credentials to authenticate your application. To use this
+method to authenticate, create a user with an associated Amazon Resource
+Name (ARN) on the ``$external`` database. Then, specify ``'MONGODB-AWS'`` as the value of
```
jmikola, line 192: I have only a basic understanding of the `MONGODB-AWS` mechanism, but I don't think this is correct. `$external` is also the default auth source used by `MONGODB-X509`, and it implies that the credentials are coming from an external source. There is no actual database named `$external`.

In fact, MongoDB doesn't allow databases to include a "$" character (see: [Naming Restrictions](https://www.mongodb.com/docs/manual/reference/limits/#naming-restrictions)).

I'll note that this text doesn't appear in the corresponding [PyMongo docs](https://www.mongodb.com/docs/languages/python/pymongo-driver/current/security/authentication/#mongodb-aws).

source/security/authentication.txt
```diff
@@ -0,0 +1,401 @@
...
+
+      .. code-block:: php
+
+         $options = [
+             'username' => '<AWS IAM access key ID>',
+             'password' => '<AWS IAM secret access key>',
+             'authMechanism' => 'MONGODB-AWS',
+         ];
+
+         $client = new Client(
```
jmikola, line 240: Same notes apply about the full class name and `$uriOptions`.

source/security/authentication.txt
```diff
@@ -0,0 +1,401 @@
...
+
+To use these environment variables to authenticate your application, first set them to the
+AWS IAM values needed for authentication, as shown in the following code
+example:
+
+.. code-block:: sh
+
+   export AWS_ACCESS_KEY_ID=<AWS IAM access key ID>
+   export AWS_SECRET_ACCESS_KEY=<AWS IAM secret access key>
+   export AWS_SESSION_TOKEN=<AWS session token>
```
jmikola, line 274: A shell script would certainly work if PHP is being launched as a CLI process, but I think this is less relevant for web applications. Most frameworks have their own method for managing env vars (e.g. [phpdotenv](https://github.com/vlucas/phpdotenv)).

[Symfony](https://symfony.com/doc/current/configuration.html#env-file-syntax) has its own implementation, and Laravel uses [phpdotenv](https://github.com/vlucas/phpdotenv) directly. The `.env` file should look the same for both frameworks:

```
AWS_ACCESS_KEY_ID=<AWS IAM access key ID>
AWS_SECRET_ACCESS_KEY=<AWS IAM secret access key>
AWS_SESSION_TOKEN=<AWS session token>
```

You can also wrap the values in double quotes if you like, although I believe that's only necessary if they contain spaces (not sure how relevant that is for these secrets).

It may make sense to present two tabs here. The shell example using `export` and a `.env` file.

source/security/authentication.txt
```diff
@@ -0,0 +1,401 @@
...
+            'mongodb://<hostname>:<port>',
+            ['authMechanism' => 'MONGODB-AWS'],
+         );
+
+   .. tab:: Connection URI
+      :tabid: connectionstring
+
+      .. code-block:: php
+
+         $uri = 'mongodb://<hostname>:<port>/?&authMechanism=MONGODB-AWS';
```
jmikola, line 309:
```suggestion
         $uri = 'mongodb://<hostname>:<port>/?authMechanism=MONGODB-AWS';
```

"&" is only a delimiter for sequential URI options. This would have been a parse error.

source/security/authentication.txt
```diff
@@ -0,0 +1,401 @@
...
+``MongoDB\Client`` constructor or through a parameter in your connection URI.
+
+.. tabs::
+
+   .. tab:: MongoDB\\Client
+      :tabid: Client
+
+      .. code-block:: php
+
+         $client = new Client(
```
jmikola, line 299: Comments about the full class name apply for both tabbed examples here.

source/security/authentication.txt
```diff
@@ -0,0 +1,401 @@
...
+         $uri = 'mongodb://<hostname>:<port>/?&authMechanism=MONGODB-AWS';
+         $client = new Client($uri);
+
+.. tip:: AWS Lambda
+   
+   AWS Lambda runtimes can automatically set these environment variables during
+   initialization. For more information about using environment variables in an AWS Lambda
+   environment, see `Using Lambda environment variables
+   <https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime>`__
+   in the AWS documentation.
```
jmikola, line 318: @GromNaN: I recall you tested the driver on Lambda, but I don't recall if it was a viable runtime (given our connection management). In your opinion, is it worth mentioning this here?

source/security/authentication.txt
```diff
@@ -0,0 +1,401 @@
...
+
+ECS Metadata
+~~~~~~~~~~~~
+
+If your application runs in an Elastic Container Service (ECS) container,
+the {+php-library+} can automatically retrieve temporary AWS credentials from an
+ECS endpoint. To do so, specify the URI of the ECS endpoint in an environment variable called
+``AWS_CONTAINER_CREDENTIALS_RELATIVE_URI``, as shown in the following example:
+
+.. code-block:: sh
```
jmikola, line 360: Consider tabbed examples for the shell and `.env` files.

source/security/authentication.txt
```diff
@@ -0,0 +1,401 @@
...
+   
+   AWS Lambda runtimes can automatically set these environment variables during
+   initialization. For more information about using environment variables in an AWS Lambda
+   environment, see `Using Lambda environment variables
+   <https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime>`__
+   in the AWS documentation.
+
+.. _php-mongodb-aws-oidc:
+
+AssumeRoleWithWebIdentity Request
```
jmikola, line 322: Note: PHP doesn't test any of this, as we rely entirely on the C driver's implementation. See [PHPC-1895](https://jira.mongodb.org/browse/PHPC-1895) for more context.

I don't consider myself qualified to review this, but if the text applies to the C driver docs it should certainly work for us. And I assume that's the case since there's no PHP code here.

In all of these AWS examples, the _only_ thing handled by PHP is constructing the client object from a connection string and any URI options. The C driver handles all reading of AWS environment variables and any API requests like `AssumeRoleWithWebIdentity`.

source/security/authentication.txt
```diff
@@ -0,0 +1,401 @@
...
+see the :ref:`php-client` guide.
+
+API Documentation
+~~~~~~~~~~~~~~~~~
+
+To learn more about the ``MongoDB\Client`` class, see :phpclass:`MongoDB\Client` 
+in the library API documentation.
+
+To view a full list of URI options that you can pass to a ``MongoDB\Client``, see
+`MongoDB\\Driver\\Manager::__construct Parameters
```
jmikola, line 399:
```suggestion
`MongoDB\\Driver\\Manager::__construct() Parameters
```

Your call, but I think the parentheses are warranted if we're referring the constructor function.

jmikola (commented):
Note that I'm a poor judge of any AWS auth docs. I've reviewed anything PHP-specific there, but have to assume anything beyond that was copied from another driver's tutorial and already vetted.

source/security/authentication.txt
norareidy: agreed

source/security/authentication.txt
```diff
@@ -0,0 +1,401 @@
...
+         $client = new Client($uri);
+
+.. _php-scram-sha-1:
+
+SCRAM-SHA-1
+-----------
+
+SCRAM-SHA-1, as defined by `RFC 5802 <https://www.rfc-editor.org/rfc/rfc5802>`__,
+is the default authentication mechanism on MongoDB deployments
+running {+mdb-server+} v3.6.
```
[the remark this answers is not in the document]
norareidy: I kept the section but added a note about dropping v3.6 support. Also removed the code example.

source/security/authentication.txt
norareidy: This is based on the [C++ page](https://www.mongodb.com/docs/languages/cpp/cpp-driver/upcoming/security/authentication/#mongodb-aws) but my understanding of AWS auth and $external is also limited. I changed the wording to something more generic, but let me know your thoughts

source/security/authentication.txt
norareidy: I added a note about this in the AWS intro section. Is it accurate to say that the C driver retrieves AWS credentials from each source in this section?

source/security/authentication.txt
```diff
@@ -0,0 +1,390 @@
...
+--------
+
+This guide describes the mechanisms you can use in the {+driver-short+} to authenticate
+users.
+
+.. important:: Percent-Encoding
+
+   You must :wikipedia:`percent-encode <Percent-encoding>` a username and password before
+   you include them in a MongoDB URI. You can use the ``rawurlencode()`` method to encode 
+   these values according to the URI syntax specified in RFC 3986. Don't percent-encode the
```
jmikola, line 30: Did you mean to use the link you defined for "RFC 3986" below?

source/security/authentication.txt
```diff
@@ -0,0 +1,390 @@
...
+   The MONGODB-AWS authentication mechanism requires {+mdb-server+} v4.4 or later.
+
+The MONGODB-AWS authentication mechanism uses AWS IAM (Amazon Web Services Identity and
+Access Management) or AWS Lambda credentials to authenticate your application. To use this
+method to authenticate, you must specify ``'MONGODB-AWS'`` as the value of
+the ``authMechanism`` connection option.
+
+.. note::
+
+   The {+php-library+} uses the C driver's implementation of the MONGODB-AWS
```
jmikola, line 156: Optional suggestion: We can refer to the C driver as "libmongoc". That may help readers since we talk about libmongoc and libmongocrypt as dependencies elsewhere in our docs, and then "driver" will only refer to PHP in these tutorials.

source/security/authentication.txt
```diff
@@ -0,0 +1,390 @@
...
+   authentication mechanism. To learn more about using this authentication mechanism
+   with the C driver, see `Authentication via AWS IAM
+   <https://www.mongodb.com/docs/languages/c/c-driver/current/libmongoc/authentication/#authentication-via-aws-iam>`__
+   in the C documentation.
+
+When you use the MONGODB-AWS mechanism, the C driver tries to retrieve AWS
+credentials from the following sources, in the order listed:
```
jmikola, line 163: I think the note above about us relying on libmongoc is sufficient, so here you can just refer to "the driver" or "the PHP driver". Going forward, I don't think we need to discuss libmongoc at all and we can just talk about obtaining these options from Client construction (be it URI string or extra options) or environment variables.

source/security/authentication.txt
jmikola: When the Manager and Client constructors receive their connection string (`$uri`) and extra options (`$uriOptions`), we pass the connection string directly to libmongoc for parsing and then iteratively apply each field in `$uriOptions` to the parsed URI object. That object is then used to create the libmongoc client.

For that reason, I wouldn't suggest any language about the "C driver" doing anything with a Client constructor. The intro section note about PHP relying on libmongoc for AWS auth mechanisms seems sufficient, and then we can just refer to the PHP driver in subsequent sections. The link to the C driver docs for AWS auth is helpful, so thanks for adding that.

source/security/authentication.txt
```diff
@@ -0,0 +1,390 @@
...
+these sources and use them to authenticate your PHP application.
+
+.. _php-mongodb-aws-credentials:
+
+MongoDB\Client Credentials
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First, the C driver checks whether you passed AWS credentials to the
+``MongoDB\Client`` constructor, either as an options parameter or as part of the
+connection URI. To pass your credentials to ``MongoDB\Client``,
```
jmikola, line 182: Consider:

> either as part of the connection string or `$uriOptions` array parameter

If "connection URI" is preferred, that's fine. I'd lead with talking about the first URI parameter, though, and then mention the options array.

source/security/authentication.txt
```diff
@@ -0,0 +1,390 @@
...
+   <https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime>`__
+   in the AWS documentation.
+
+.. _php-mongodb-aws-oidc:
+
+AssumeRoleWithWebIdentity Request
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If your application authenticates users for your EKS cluster from an OpenID Connect (OIDC)
+identity provider, the C driver can make an ``AssumeRoleWithWebIdentity`` request
```
jmikola, line 301: Here I would just refer to "the driver", as the user doesn't really need to think about whether PHP or libmongoc is doing the heavy lifting. The PHP extension is a black box for them.

My earlier explanation about libmongoc and PHP was more to provide you context, and explain why we (the PHP team) don't have much knowledge about the AWS mechanisms.

source/security/authentication.txt
```diff
@@ -0,0 +1,390 @@
...
+   - `AssumeRoleWithWebIdentity
+     <https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html>`__
+
+.. _php-mongodb-aws-ecs:
+
+ECS Metadata
+~~~~~~~~~~~~
+
+If your application runs in an Elastic Container Service (ECS) container,
+the C driver can automatically retrieve temporary AWS credentials from an
```
jmikola, line 331:
```suggestion
the driver can automatically retrieve temporary AWS credentials from an
```

source/security/authentication.txt
```diff
@@ -0,0 +1,390 @@
...
+connection option to ``'MONGODB-AWS'``. To view an example that sets the
+``authMechanism`` option, see the :ref:`authMechanism example
+<php-mongodb-aws-env-example>` on this page.
+
+.. _php-mongo-aws-ec2:
+
+EC2 Instance Metadata
+~~~~~~~~~~~~~~~~~~~~~
+
+The C driver can automatically retrieve temporary AWS credentials from an
```
jmikola, line 363:
```suggestion
The driver can automatically retrieve temporary AWS credentials from an
```

source/security/authentication.txt
norareidy: We try to avoid inline links that aren't introduced, and since there's already one to Wikipedia I just linked in the list below

jmikola: Oh, I missed that those were part of a list and assumed they were internal references that didn't get used. Disregard my earlier comment.

source/security/authentication.txt
jmikola: The C driver page inconsistently says "`$external` virtual database" and "`$external` database". I think including "virtual" makes sense.

I see you removed all mention of "$external" in this tutorial, which seems ideal.

jmikola (approved):

PR #134, opened 2024-09-10
DOCSP-41984: single field index

rustagir:
# Pull Request Info

[PR Reviewing Guidelines](https://github.com/mongodb/docs-php-library/blob/master/REVIEWING.md)

JIRA - <https://jira.mongodb.org/browse/DOCSP-41984>
Staging:
- [index mgmt](https://preview-mongodbrustagir.gatsbyjs.io/php-library/DOCSP-41984-single-fld-idx/indexes/index-mgmt/)
- [single field indexes](https://preview-mongodbrustagir.gatsbyjs.io/php-library/DOCSP-41984-single-fld-idx/indexes/single-field-index/)

## Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST? all errors unrelated
- [x] Did you run a spell-check?
- [x] Did you run a grammar-check?
- [x] Are all the links working?
- [x] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?

PR #140, opened 2024-09-12
DOCSP-41986: multikey indexes

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-41986>
Staging - https://preview-mongodbrustagir.gatsbyjs.io/php-library/DOCSP-41986-multikey-indexes/indexes/multikey-index/

## Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST? Unrelated to PR

### DIFF
--- a/source/includes/indexes/indexes.php
+++ b/source/includes/indexes/indexes.php
@@ -30,3 +30,14 @@ $indexName = $collection->createIndex(['title' => 1]);
 $document = $collection->findOne(['title' => 'Sweethearts']);
 echo json_encode($document) , PHP_EOL;
 // end-index-single-query
+
+// start-multikey
+$indexName = $collection->createIndex(['cast' => 1]);
+// end-multikey
+
+// start-index-array-query
+$document = $collection->findOne(
+    ['cast' => ['$in' => ['Aamir Khan', 'Kajol']]]
+);
+echo json_encode($document) , PHP_EOL;
+// end-index-array-query
--- a/source/includes/indexes/indexes.php
+++ b/source/includes/indexes/indexes.php
@@ -28,7 +28,7 @@ $indexName = $collection->createIndex(['title' => 1]);
 
 // start-index-single-query
 $document = $collection->findOne(['title' => 'Sweethearts']);
-echo json_encode($document) , PHP_EOL;
+echo json_encode($document), PHP_EOL;
 // end-index-single-query
 
 // start-multikey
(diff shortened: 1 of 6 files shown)


source/indexes/multikey-index.txt
```diff
@@ -0,0 +1,85 @@
...
+.. meta::
+   :keywords: index, query, optimization, efficiency
+
+Overview
+--------
+
+**Multikey indexes** are indexes that improve the performance of queries
+on array-valued fields. You can create a multikey index on a collection
+by using the ``MongoDB\Collection::createIndex()`` method and the same
+syntax that you use to create single field or compound indexes.
```
jordan-smith721, line 26: S: Might be helpful to link to the single field and compound index pages as you mention them here

jordan-smith721 (approved):
One non-blocking suggestion but LGTM!

source/includes/indexes/indexes.php
```diff
@@ -30,3 +30,14 @@
...
+
+// start-multikey
+$indexName = $collection->createIndex(['cast' => 1]);
+// end-multikey
+
+// start-index-array-query
+$document = $collection->findOne(
+    ['cast' => ['$in' => ['Aamir Khan', 'Kajol']]]
+);
+echo json_encode($document) , PHP_EOL;
```
jmikola, line 42: In the index landing PR (https://github.com/mongodb/docs-php-library/pull/127), I suggested using some driver functions to print the value as extended JSON. It looks like that was missed in a subsequent PR for single field indexes (https://github.com/mongodb/docs-php-library/pull/134), upon which I expect this was based.

I'll defer to you if it's worth changing. The expected output in the RST doesn't show any special BSON types, so it's probably of little consequence.

```suggestion
echo json_encode($document), PHP_EOL;
```

No need for preceding whitespace for a comma. It would be appropriate for a concatenation operator (dot), though.

jmikola (approved):
LGTM with suggested whitespace change.

If looks like that was also missed in https://github.com/mongodb/docs-php-library/pull/134 so you can change the other `json_encode()` line as well for consistency.

source/includes/indexes/indexes.php
rustagir: I think it's ok to use the simpler `json_encode()` method in these contexts. I will change the whitespace though!

PR #144, opened 2024-09-17
DOCSP-41982: cluster monitoring

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-41982>
Staging - https://deploy-preview-144--docs-php-library.netlify.app/monitoring/cluster-monitoring/

## Self-Review Checklist


source/monitoring.txt
```diff
@@ -0,0 +1,21 @@
...
+
+.. toctree::
+   :caption: Monitoring categories
+
+   /monitoring/cluster-monitoring
+
+.. /monitoring/command-monitoring
+.. /monitoring/connection-monitoring
+
+- :ref:`Cluster Monitoring <php-cluster-monitoring>`: monitor changes
```
mongoKart, line 15: I: https://www.mongodb.com/docs/meta/style-guide/style/punctuation/#colons

```suggestion
- :ref:`Cluster Monitoring <php-cluster-monitoring>`: Monitor changes
```

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,135 @@
...
+   :depth: 2
+   :class: singlecols
+
+Overview
+--------
+
+This guide shows you how to use the {+php-library+} to monitor topology
+events in a MongoDB instance, replica set, or sharded cluster. The
+driver creates topology events, also known as Server Discovery and
+Monitoring (**SDAM**) events, when there are any changes in the state of the
```
mongoKart, line 26: S: i know SDAM is the more commonly used term, but i wonder if bolding the whole thing makes more sense

```suggestion
driver creates topology events, also known as **Server Discovery and
Monitoring (SDAM)** events, when there are any changes in the state of the
```

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,135 @@
...
+
+You might use information about topology events in your
+application to understand cluster changes, assess cluster health, or
+perform capacity planning.
+
+Subscribe to Events
+-------------------
+
+You can access details about SDAM events by subscribing to them
+in your application. To subscribe to any event, create a class that
```
mongoKart, line 37: S:

```suggestion
in your application. To subscribe to an event, create a class that
```

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,135 @@
...
+implements the ``MongoDB\Driver\Monitoring\SDAMSubscriber`` interface,
+then use the ``MongoDB\Client::addSubscriber()`` method to register the
+event subscriber with your ``MongoDB\Client`` instance.
+
+The following code creates the ``MySubscriber`` class that implements
+``SDAMSubscriber`` to output a message when a ``ServerOpeningEvent`` is
```
mongoKart, line 43: I: https://www.mongodb.com/docs/meta/style-guide/writing/use-that-which-correctly/

```suggestion
The following code creates the ``MySubscriber`` class, which implements
``SDAMSubscriber`` to output a message when a ``ServerOpeningEvent`` is
```

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,135 @@
...
+   :start-after: start-mysubscriber
+   :end-before: end-mysubscriber
+   :language: php
+   :copyable:
+   :dedent:
+
+.. note::
+
+   As shown in the preceding code, you must implement all of the methods
+   of ``SDAMSubscriber``, even for events you are not subscribing to.
```
mongoKart, line 56: S:

```suggestion
   of the ``SDAMSubscriber`` interface, even for events you are not subscribing to.
```

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,135 @@
...
+   You can implement these methods with empty bodies so that the library
+   does not generate any messages for these events.
+
+Then, the following code uses the ``addSubscriber()`` method to register
+``MySubscriber`` with the client:
```
mongoKart, line 61: I: I like the use of "then" to signify a subsequent step, but I would tie it to a user step instead of the code example. Something like:

```suggestion
Then, use the ``addSubscriber()`` method to register
``MySubscriber`` with the client, as shown in the following code:
```

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,135 @@
...
+
+.. list-table::
+   :widths: 35 65 
+   :header-rows: 1
+
+   * - Event Name
+     - Description
+
+   * - ``ServerChangedEvent``
+     - Created when an instance state changes, such as from secondary to
```
mongoKart, line 94: Optional: Could remove "Created when" from the start of every event (probably along with a new column header name)

mongoKart (changes requested):
Looks really good! A few very small changes

source/monitoring/cluster-monitoring.txt
rustagir: probably ok to go without any bolding, tbh

source/monitoring/cluster-monitoring.txt
rustagir: reworded

source/monitoring/cluster-monitoring.txt
rustagir: I can't think of a good way to adjust the column header, as renaming it "Situation that Triggers Event" seems worse than keeping the "Created when"s in

mongoKart (approved):
LGTM

source/includes/monitoring/sdam.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+require __DIR__ . '/vendor/autoload.php';
+
+// start-mysubscriber
+class MySubscriber implements MongoDB\Driver\Monitoring\SDAMSubscriber
+{
+    private $stream;
+    public function __construct($stream)
```
jmikola, line 9:
```suggestion

    public function __construct($stream)
```

For readability, I'd suggest line breaks between a group of property declarations and any methods.

source/includes/monitoring/sdam.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        $this->stream = $stream;
+    }
+    public function serverOpening(MongoDB\Driver\Monitoring\ServerOpeningEvent $event): void
```
jmikola, line 13:
```suggestion

    public function serverOpening(MongoDB\Driver\Monitoring\ServerOpeningEvent $event): void
```

Also, between any methods with bodies.

source/includes/monitoring/sdam.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        fwrite($this->stream, sprintf(
+            'Server opening on %s:%s%s',
+            $event->getHost(),
+            $event->getPort(),
+            PHP_EOL,
+        ));
+    }
+    public function serverClosed(MongoDB\Driver\Monitoring\ServerClosedEvent $event): void {}
```
jmikola, line 22:
```suggestion

    public function serverClosed(MongoDB\Driver\Monitoring\ServerClosedEvent $event): void {}
```

Since the methods below have no bodies, I'm fine grouping them together.

source/includes/monitoring/sdam.php
```diff
@@ -0,0 +1,44 @@
...
+    public function serverChanged(MongoDB\Driver\Monitoring\ServerChangedEvent $event): void {}
+    public function serverHeartbeatFailed(MongoDB\Driver\Monitoring\ServerHeartbeatFailedEvent $event): void {}
+    public function serverHeartbeatStarted(MongoDB\Driver\Monitoring\ServerHeartbeatStartedEvent $event): void {}
+    public function serverHeartbeatSucceeded(MongoDB\Driver\Monitoring\ServerHeartbeatSucceededEvent $event): void {}
+    public function topologyChanged(MongoDB\Driver\Monitoring\TopologyChangedEvent $event): void {}
+    public function topologyClosed(MongoDB\Driver\Monitoring\TopologyClosedEvent $event): void {}
+    public function topologyOpening(MongoDB\Driver\Monitoring\TopologyOpeningEvent $event): void {}
```
jmikola, line 29: I believe [contravariance](https://www.php.net/manual/en/language.oop5.variance.php#language.oop5.variance.contravariance) would allow us to remove the type hints from these methods and make the code more concise. See: https://3v4l.org/H34oT#v8.3.11

That said, it'd make the example code less useful for copying if users want to implement these methods themselves. I'm fine to leave these as-is.

/cc @alcaeus

source/includes/monitoring/sdam.php
```diff
@@ -0,0 +1,44 @@
...
+}
+// end-mysubscriber
+
+$uri = getenv('MONGODB_URI') ?: throw new RuntimeException('Set the MONGODB_URI variable to your Atlas URI that connects to the sample dataset');
```
jmikola, line 33: "that connects to the sample dataset" doesn't seem relevant here. @norareidy changed this in a previous PR where I pointed that out, if you want to make it consistent.

source/includes/monitoring/sdam.php
```diff
@@ -0,0 +1,44 @@
...
+$client = new MongoDB\Client($uri);
+
+$database = $client->db;
+$collection = $database->my_coll;
```
jmikola, line 37:
```suggestion
$collection = $client->db->my_coll;
```

`$database` is never used.

source/includes/monitoring/sdam.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->stream = $stream;
+    }
+    public function serverOpening(MongoDB\Driver\Monitoring\ServerOpeningEvent $event): void
+    {
+        fwrite($this->stream, sprintf(
+            'Server opening on %s:%s%s',
+            $event->getHost(),
+            $event->getPort(),
+            PHP_EOL,
+        ));
```
jmikola, line 20:
```suggestion
        fprintf(
            $this->stream,
            "Server opening on %s:%s\n",
            $event->getHost(),
            $event->getPort(),
        );
```

See: https://www.php.net/fprintf

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,136 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecols
+
+Overview
+--------
+
+This guide shows you how to use the {+php-library+} to monitor topology
+events in a MongoDB instance, replica set, or sharded cluster. The
+driver creates topology events, also known as Server Discovery and
```
jmikola, line 25: Technically, topology events are those related to the entire cluster, while server events are those pertaining to a single server within the topology. [SDAM monitoring](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.md) includes both types of events.

I would personally just refer to them as SDAM events, as you seem to do in a subsequent section. If you need a short name, "cluster events" might also work but there's no prior art for that in any driver spec.

Disregard if this name was intentionally chosen for consistency with existing driver docs.

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,136 @@
...
+   :language: php
+   :copyable:
+   :dedent:
+
+.. note::
+
+   As shown in the preceding code, you must implement all of the methods
+   of the ``SDAMSubscriber`` interface, even for events you are not subscribing to.
+   You can implement these methods with empty bodies so that the library
+   does not generate any messages for these events.
```
jmikola, line 59: Technically, the library never "generate(s) any messages". That's something you've chosen to do in the subscriber.

This would be more accurate (feel free to reword):

> This example defines the additional methods as empty so not to output any messages for those events.

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,136 @@
...
+
+   Server opening on ac-rmuag0v-shard-00-00.gh0qg50.mongodb.net:27017
+   Server opening on ac-rmuag0v-shard-00-01.gh0qg50.mongodb.net:27017
+   Server opening on ac-rmuag0v-shard-00-02.gh0qg50.mongodb.net:27017
+
+Event Descriptions
+------------------
+
+You can subscribe to the following SDAM events by implementing the
+corresponding method from the ``SDAMSubscriber`` interface:
```
jmikola, line 85: This talks about the "corresponding method" from [SDAMSubscriber](https://www.php.net/manual/en/class.mongodb-driver-monitoring-sdamsubscriber.php), but those names aren't presented in the table below. It would be a bit redundant to dedicate a column to the method name, but I think some links to the PHP.net docs could make this a more helpful resource.

Consider working in links to the following:

 * The [SDAMSubscriber](https://www.php.net/manual/en/class.mongodb-driver-monitoring-sdamsubscriber.php) interface.
 * Individual class pages for each event under https://www.php.net/manual/en/mongodb.monitoring.php. For example, ServerOpeningEvent can link to https://www.php.net/manual/en/class.mongodb-driver-monitoring-serveropeningevent.php.

source/monitoring/cluster-monitoring.txt
jmikola: In the docs for [ServerChangedEvent](https://www.php.net/manual/en/class.mongodb-driver-monitoring-serverchangedevent.php) we say the event class "encapsulates information about a changed server description".

Here, the table is presenting the event as the thing that happened instead of just an object containing some information about a thing that happened.

The "Condition" column in [Events API](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.md#events-api-1) from the driver spec might be helpful. Note that the spec includes "Description" in the name of these events, where PHP just calls it ServerChangedEvent. I think that was due to us copying the shorter name from our libmongoc dependency.

I might suggest:

> When the server description changes, such as the server's type changing from secondary to primary.

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,136 @@
...
+     - Created when the heartbeat fails.
+
+API Documentation
+-----------------
+
+To learn more about any of the classes or methods discussed in this guide, see the
+following API documentation:
+
+- :phpmethod:`MongoDB\Client::addSubscriber()`
+- :phpclass:`MongoDB\Client`
```
jmikola, line 130: The `removeSubscriber()` method would be a more helpful reference. I don't see any reason to link to the Client class page directly.

On that note, perhaps you should include a brief example of unregistering a subscriber.

source/includes/monitoring/sdam.php
rustagir: I'll make the change to the message!

source/includes/monitoring/sdam.php
rustagir: Good to know! I grabbed that code from here intially: https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBClient-addSubscriber/#example

source/monitoring/cluster-monitoring.txt
rustagir: Good to know, in general it is ok if this page differs from the same page in other docs. Each docs standardization is meant to improve on previous ones!

source/monitoring/cluster-monitoring.txt
rustagir: The API doc for SDAMSubscriber is linked at the end of the page, but I will link each event in the table.

source/includes/monitoring/sdam.php
alcaeus: Omitting the types would make the example more concise, but I'd like to avoid suggesting to do this in documentation, as I consider it bad practice. We could leverage `use` statements to shorten the types, but that may be difficult if only small snippets of this code are included in the documentation.

We face a similar issue with the other documentation examples used in the server docs, and we decided to just deal with the long types in favour of having copy-pastable code.

source/includes/monitoring/sdam.php
jmikola: Looks like @GromNaN originally wrote that in https://github.com/mongodb/mongo-php-library/commit/b455ab7c20ce337653efe39afed19b8a37c0ff5a :smile:

source/monitoring/cluster-monitoring.txt
```diff
@@ -81,53 +83,79 @@ outputs messages such as the following:
...
 
-   * - ``ServerOpeningEvent``
-     - Created when the server is initialized.
+   * - :php:`ServerOpeningEvent <mongodb-driver-monitoring-serveropeningevent>`
+     - Created when the server description is instantiated with its
+       defaults.
 
-   * - ``ServerClosedEvent``
+   * - :php:`ServerClosedEvent <mongodb-driver-monitoring-serverclosedevent>`
      - Created when the server is closed.
```
jmikola, line 107: This can say "when a server connection is closed". For other drivers, this only pertains to the monitoring thread's connections, but monitoring and app connections are one and the same in PHP since our SDAM implementation is single-threaded.

source/monitoring/cluster-monitoring.txt
```diff
@@ -81,53 +83,79 @@ outputs messages such as the following:
...
-     - Created when an instance state changes, such as from secondary to
-       primary.
+   * - :php:`ServerChangedEvent <mongodb-driver-monitoring-serverchangedevent>`
+     - Created when the server description changes, such as the server's type
+       changing from secondary to primary.
 
-   * - ``ServerOpeningEvent``
-     - Created when the server is initialized.
+   * - :php:`ServerOpeningEvent <mongodb-driver-monitoring-serveropeningevent>`
+     - Created when the server description is instantiated with its
```
jmikola, line 103: For PHP, I think this can say "when a server connection is established". I'm asking for feedback on this in the drivers channel, but I don't think the spec's language about instantiating a server description is relevant to users.

For other drivers, I expect the public-facing docs would be "when the driver establishes a monitoring connection to the server".

Depending on the feedback I receive, perhaps we may want to say "immediately before the driver establishes a connection to the server" (or "monitoring connection to the server" for other drivers).

source/monitoring/cluster-monitoring.txt
```diff
@@ -81,53 +83,79 @@ outputs messages such as the following:
...
+methods in the :php:`Monitoring classes and subscriber functions
+<mongodb.monitoring>` section of the PHP manual.
+
+Remove a Subscriber
+-------------------
+
+Later in your application, you might not want to subscribe to
+SDAM events. To unregister a subscriber from your client, use the
+``MongoDB\Client::removeSubscriber()`` method. If you attempt to remove
+a nonexistent subscriber, the method doesn't perform any action.
```
jmikola, line 139: On a related note, PHPC (extension) also doesn't allow you to register the same subscriber twice. Attempting to do so would also be a no-op. I don't think we need to mention that, though.

jmikola (commented):
Some suggested rewording, but I want to wait until I hear back from other folks in drivers before signing off on the ServerOpeningEvent text.

source/monitoring/cluster-monitoring.txt
jmikola: @rustagir: Feedback came back. I think we're good to simply say "when a server connection is established".

source/monitoring/cluster-monitoring.txt
```diff
@@ -81,53 +83,79 @@ outputs messages such as the following:
...
+       an election of a new primary or disconnection of a ``mongos`` proxy.
 
-   * - ``TopologyOpeningEvent``
-     - Created when the topology is initialized.
+   * - :php:`TopologyOpeningEvent <mongodb-driver-monitoring-topologyopeningevent>`
+     - Created when the topology description is initialized.
 
-   * - ``TopologyClosedEvent``
+   * - :php:`TopologyClosedEvent <mongodb-driver-monitoring-topologyclosedevent>`
      - Created when the topology is closed.
```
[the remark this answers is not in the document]
rustagir: thanks for providing this background!

source/monitoring/cluster-monitoring.txt
```diff
@@ -81,53 +83,79 @@ outputs messages such as the following:
...
-   * - ``ServerClosedEvent``
+   * - :php:`ServerClosedEvent <mongodb-driver-monitoring-serverclosedevent>`
      - Created when the server is closed.
 
-   * - ``TopologyChangedEvent``
-     - Created when the topology changes, such as an election of a new
-       primary or disconnection of a ``mongos`` proxy.
+   * - :php:`TopologyChangedEvent <mongodb-driver-monitoring-topologychangedevent>`
+     - Created when the topology description changes, such when there is
+       an election of a new primary or disconnection of a ``mongos`` proxy.
```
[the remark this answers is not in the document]
rustagir: Ah, that must be a copy paste artifact.

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,163 @@
...
+       an election of a new primary.
+
+   * - :php:`TopologyOpeningEvent <mongodb-driver-monitoring-topologyopeningevent>`
+     - Created when the driver first connects to the cluster.
+
+   * - :php:`TopologyClosedEvent <mongodb-driver-monitoring-topologyclosedevent>`
+     - Created when the driver disconnects from the cluster.
+
+   * - :php:`ServerHeartbeatStartedEvent <mongodb-driver-monitoring-serverheartbeatstartedevent>`
+     - Created when the server monitor sends a ``hello`` call to the server.
```
jmikola, line 119:
```suggestion
     - Created when the server monitor sends a ``hello`` command to the server.
```

"Command" is less ambiguous.

source/monitoring/cluster-monitoring.txt
```diff
@@ -0,0 +1,163 @@
...
+       changing from secondary to primary.
+
+   * - :php:`ServerOpeningEvent <mongodb-driver-monitoring-serveropeningevent>`
+     - Created when a server connection is established.
+
+   * - :php:`ServerClosedEvent <mongodb-driver-monitoring-serverclosedevent>`
+     - Created when a server connection is closed.
+
+   * - :php:`TopologyChangedEvent <mongodb-driver-monitoring-topologychangedevent>`
+     - Created when the topology description changes, such when there is
```
jmikola, line 109: I think you accidentally a word in this sentence.

```suggestion
     - Created when the topology description changes, such as when there is
```

jmikola (approved):
LGTM with my suggestions applied.

PR #152, opened 2024-09-25
DOCSP-41992 Upgrade versions

lindseymoore:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-41992>
Staging - <https://deploy-preview-152--docs-php-library.netlify.app/upgrade/>

## Self-Review Checklist

### DIFF
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -1,4 +1,4 @@
-.. php-upgrade:
+.. _php-upgrade:
 
 =======================
 Upgrade Driver Versions
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -58,7 +58,7 @@ directory:
 
    pecl upgrade mongodb-<version-number>
 
-To upgrade the PHP Library version, use the following command:
+To upgrade the PHP library version, run the following command:
 
 .. code-block:: bash
 
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -20,9 +20,9 @@ Upgrade Driver Versions
 Overview
 --------
 
-This page describes how to upgrade your driver to a new version and the changes 
-you must make to your application when you upgrade to a new version of the 
-{+driver-short+}.
+On this page, you can learn how to upgrade your driver to a new version. This
+page also includes the changes you must make to your application when you 
+upgrade to a new version of the {+driver-short+}.
 
 How to Upgrade
 --------------
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -43,16 +43,16 @@ Before you upgrade, perform the following actions:
    To ensure compatibility across {+mdb-server+} versions when
    upgrading driver versions, use the :ref:`{+stable-api+} <php-stable-api>`.
 
-To upgrade your driver version, replace "version number" with the version number 
+To upgrade your driver version, replace ``<version-number>`` with the version number 
 you want to upgrade to and run the following command in your application's 
 directory:
 
 .. code-block:: bash
 
-   composer require mongodb/mongodb:<version number>
+   composer require mongodb/mongodb:<version-number>
 
 Installation instructions for the ``mongodb`` extension may be found in the 
-`PHP.net documentation <https://www.php.net/manual/en/mongodb.installation.php>`__.
+:php:`PHP.net documentation <manual/en/mongodb.installation.php>`.
 
 .. _php-breaking-changes:
 
@@ -77,14 +77,14 @@ This driver version introduces the following breaking changes:
 
 - Drops support for {+mdb-server+} 3.6.
 
-- Deprecate iterators for database, collection, and index enumeration.
+- Deprecates iterators for database, collection, and index enumeration.
 
 Version 1.18 Breaking Changes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 This driver version introduces the following breaking changes:
 
-- Deprecate setting ``disableMD5`` to ``false`` when instantiating a Bucket due to 
+- Deprecates setting ``disableMD5`` to ``false`` when instantiating a Bucket due to 
   deprecated MD5 encoding. The default value for the setting is now ``true``. 
 
 Version 1.17 Breaking Changes
@@ -96,7 +96,7 @@ This driver version introduces the following breaking changes:
 
 - Requires PHP 7.4 or newer.
 
-- Remove aggregate compatibility logic, like the ``useCursor`` option, for 
+- Removes aggregate compatibility logic, like the ``useCursor`` option, for 
   {+mdb-server+} versions less than 3.6.
 
 Version 1.16 Breaking Changes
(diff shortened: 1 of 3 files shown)


source/upgrade.txt
```diff
@@ -0,0 +1,123 @@
...
+- Address any breaking changes between the driver version
+  your application is using and your planned upgrade version in the
+  :ref:`Breaking Changes <php-breaking-changes>` section.
+
+.. tip::
+
+   To ensure compatibility across {+mdb-server+} versions when
+   upgrading driver versions, use the :ref:`{+stable-api+} <php-stable-api>`.
+
+To upgrade your driver version, replace "version number" with the version number
```
norareidy, line 46: S: use the placeholder syntax
```suggestion
To upgrade your driver version, replace ``<version-number>`` with the version number 
```

source/upgrade.txt
```diff
@@ -0,0 +1,123 @@
...
+To upgrade your driver version, replace "version number" with the version number 
+you want to upgrade to and run the following command in your application's 
+directory:
+
+.. code-block:: bash
+
+   composer require mongodb/mongodb:<version number>
+
+Installation instructions for the ``mongodb`` extension may be found in the 
+`PHP.net documentation <https://www.php.net/manual/en/mongodb.installation.php>`__.
```
norareidy, line 55: S: there's a :php: directiven you can use here:
```suggestion
:php:`PHP.net documentation <mongodb.installation>`.
```

source/upgrade.txt
```diff
@@ -0,0 +1,123 @@
...
+JIRA tickets for each release on `GitHub <https://github.com/mongodb/mongo-php-library/releases>`__.
+
+Version 1.20 Breaking Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This driver version introduces the following breaking changes:
+
+- Drops support for {+mdb-server+} 3.6.
+
+- Deprecate iterators for database, collection, and index enumeration.
```
norareidy, line 80: S: match the verb tense of the last bullet
```suggestion
- Deprecates iterators for database, collection, and index enumeration.
```

source/upgrade.txt
```diff
@@ -0,0 +1,123 @@
...
+
+Version 1.18 Breaking Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This driver version introduces the following breaking changes:
+
+- Deprecate setting ``disableMD5`` to ``false`` when instantiating a Bucket due to
```
norareidy, line 87: S:
```suggestion
- Deprecates setting ``disableMD5`` to ``false`` when instantiating a Bucket due to 
```

source/upgrade.txt
```diff
@@ -0,0 +1,123 @@
...
+Version 1.17 Breaking Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This driver version introduces the following breaking changes:
+
+- Upgrades the ``mongodb`` extension requirement to 1.17.0.
+
+- Requires PHP 7.4 or newer.
+
+- Remove aggregate compatibility logic, like the ``useCursor`` option, for
```
norareidy, line 99: S:
```suggestion
- Removes aggregate compatibility logic, like the ``useCursor`` option, for 
```

source/upgrade.txt
```diff
@@ -0,0 +1,123 @@
...
+  {+mdb-server+} versions less than 3.6.
+
+Version 1.16 Breaking Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This driver version introduces the following breaking changes:
+
+- Upgrades the ``mongodb`` extension requirement to 1.16.0.
+
+- Deprecate ``IndexInfo::isGeoHaystack()``. ``geoHaystack`` was removed in
```
norareidy, line 109: S:
```suggestion
- Deprecates ``IndexInfo::isGeoHaystack()``. ``geoHaystack`` was removed in
```

norareidy: Applies to the three bullet points below too

norareidy (approved):
LGTM with some minor suggestions!

source/upgrade.txt
```diff
@@ -0,0 +1,123 @@
...
+To upgrade your driver version, replace ``<version-number>`` with the version number 
+you want to upgrade to and run the following command in your application's 
+directory:
+
+.. code-block:: bash
+
+   composer require mongodb/mongodb:<version-number>
+
+Installation instructions for the ``mongodb`` extension may be found in the 
+:php:`PHP.net documentation <manual/en/mongodb.installation.php>`.
```
alcaeus, line 55: We should actually list the instructions on how to install the extension as well. This can be done by running the following:
```
pecl upgrade mongodb-<version-number>
```

The note about the PHP documentation can be amended to read "Detailed installation instructions [...]"

We should highlight that while the major and minor version numbers of the extension and library are in sync, the patch version is not.

source/upgrade.txt
```diff
@@ -0,0 +1,123 @@
...
+
+.. _php-breaking-changes:
+
+Breaking Changes
```
alcaeus, line 59: A couple of notes here:
* We don't consider dropping support for a server version a breaking change, otherwise it would only be done in a major version. When we drop support for a server version, it has been out of official support for several years.
* Deprecations are not considered breaking changes - rather is it a way to announce an upcoming breaking change that will be made in a future major version.
* Upgrading the extension requirement or PHP requirements are also not breaking changes, as composer's dependency resolution will refuse to install a version of the library that isn't compatible with the installed extension version or the installed PHP version.

To summarise, this section should mention that we will make breaking changes only in a new major version, and we're also providing an upgrade file to go along with that version, along with deprecating any functionality to be removed. For example, we're currently working on version 2.0 of the extension, and the [corresponding upgrade file](https://github.com/mongodb/mongo-php-driver/blob/v2.x/UPGRADE-2.0.md) lists everything that is removed. Note that this is a work in progress and should not yet be added to the docs.

We can list dropping support for MongoDB 3.6 in version 1.20 as a breaking change, but the other changes listed are not breaking changes in a traditional sense and should be removed.

alcaeus (changes requested):

source/upgrade.txt
lindseymoore: I will remove all the deprecation notes other than the dropped support for MongoDB Server 3.6 for Version 1.20. 
Once version 2.0 is released, we can add a note about upgrade files, as there are no other upgrade files for current versions out (the last one I see was for upgrading to 1.5 which is now the oldest relevant version according to compatibility table). Let me know know if this looks good, thanks!

source/upgrade.txt
lindseymoore: Added explanation about upgrading for different versions. Let me know what you think. thanks!

alcaeus (approved):

PR #153, opened 2024-09-26
Revise descriptions for server opening/closed events

jmikola:
After [DOCSP-41982](https://jira.mongodb.org/browse/DOCSP-41982) was completed, I opened [PHPC-2449](https://jira.mongodb.org/browse/PHPC-2449) to improve the PHP.net docs. While doing so, I realized the original descriptions for these events were inaccurate.

This is based on changes in https://github.com/php/doc-en/pull/3802

Note that there are additional caveats for the topology open/closed events (the driver's client persistence behavior may get in the way of observing them), but I don't think we need to discuss that in this tutorial.


rustagir (approved):
lgtm - I can merge! thanks for contributing

PR #206, opened 2025-02-18
DOCSP-43464: operations with builders

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-43464>
Staging - https://deploy-preview-206--docs-php-library.netlify.app/builders/

## Self-Review Checklist

### DIFF
--- a/source/builders.txt
+++ b/source/builders.txt
@@ -23,19 +23,21 @@ Overview
 In this guide, you can learn about the **builder classes**
 that the {+library-short+} provides to create types used in your
 operations. You can use the builder classes and factory methods from the
-:ref:`php-aggregation-builder-api` feature to create filters for other
-operations such as find, update, and delete operations.
+Aggregation Builder feature to create filters for other operations such
+as find, update, and delete operations. To learn more about the Aggregation
+Builder, see the :ref:`php-aggregation-builder-api` section of the
+Aggregation guide. 
 
 Using builders to create queries helps you identify errors at compile
 time and avoid them at runtime. This guide provides information on
-builder classes that you can use for the following tasks:
+builder classes that you can use to perform the following tasks:
 
-- :ref:`Creating a filter definition <php-builders-filter>`
-- :ref:`Defining an update operation <php-builders-update>`
-- :ref:`Filtering a change stream <php-builders-changestream>`
+- :ref:`php-builders-filter`
+- :ref:`php-builders-update`
+- :ref:`php-builders-changestream`
 
-.. note:: Options
-    
+.. note:: Setting Operation Options
+
    You cannot specify options by using factory methods for the equivalent
    aggregation stages. For example, you cannot use the ``Stage::limit()`` method
    to set a returned documents limit on your find operation. You must specify
@@ -97,7 +99,7 @@ builders:
 1. Instantiate the ``QueryInterface`` by calling the ``Query::query()``
    method.
    
-#. Pass the field name to compare and a factory method from the
+#. Pass the field name to filter on and a factory method from the
    ``Query`` class. You can pass one or more pairs of field names and
    criteria to the ``QueryInterface``.
 
@@ -187,8 +189,8 @@ You can use factory methods from the ``Stage`` builder class to update
 or create field values in your update operations. To learn more about
 updating documents, see the :ref:`php-write-update` guide.
 
-.. important::
-    
+.. important:: Only $set Operator Supported
+
    The only update operation that you can express by using builders uses
    the ``$set`` operator. This operator allows you to replace the value of
    a field with a specified value or create a new field that has a specified
@@ -218,8 +220,7 @@ This example performs the following actions:
 
 - Uses the ``Query::eq()`` factory method to match documents in which
   the ``watlev`` field value is ``'partly submerged at high water'``
-- Uses builders to specify an update that sets
-  sets the ``year`` field to ``1870``
+- Uses the ``Stage::set()`` method to update the ``year`` field to ``1870``
 - Calls the :phpmethod:`MongoDB\Collection::updateOne()`
   method to perform the update
 - Prints the number of updated documents
@@ -280,8 +281,8 @@ This example performs the following actions:
 - Uses the ``Stage::project()`` method to output only the specified change
   event fields
 - Calls the :phpmethod:`MongoDB\Collection::watch()`
-  method to open the change stream with the option to output the full
-  document after update
+  method to open the change stream and sets the ``fullDocument`` option
+  to output the full document after update
 - Prints change events as they occur
 
 .. io-code-block::
--- a/source/builders.txt
+++ b/source/builders.txt
@@ -260,12 +260,12 @@ The following code shows the template to modify change stream output by
 using builders:
 
 .. code-block:: php
-    
-   $pipeline = new Pipeline(
+
+   $pipeline = [
        Stage::<factory method>(...),
        Stage::<factory method>(...),
        ...
-   );
+   ];
 
 You can pass this pipeline to the following methods:
 
--- a/source/builders.txt
+++ b/source/builders.txt
@@ -1,0 +1,304 @@
+.. _php-builders:
+
+========================
+Operations with Builders
+========================
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. facet::
+   :name: genre
+   :values: reference
+
+.. meta::
+   :keywords: aggregation, query, code example, type-safe
+
+Overview
+--------
+
+In this guide, you can learn about the **builder classes**
+that the {+library-short+} provides to create types used in your
+operations. You can use the builder classes and factory methods from the
+:ref:`php-aggregation-builder-api` feature to create filters for other
+operations such as find, update, and delete operations.
+
+Using builders to create queries helps you identify errors at compile
+time and avoid them at runtime. This guide provides information on
+builder classes that you can use for the following tasks:
+
+- :ref:`Creating a filter definition <php-builders-filter>`
+- :ref:`Defining an update operation <php-builders-update>`
+- :ref:`Filtering a change stream <php-builders-changestream>`
+
+.. note:: Options
+    
+   You cannot specify options by using factory methods for the equivalent
+   aggregation stages. For example, you cannot use the ``Stage::limit()`` method
+   to set a returned documents limit on your find operation. You must specify
+   options by using the string-based syntax, as shown in the following code:
+   
+   .. code-block:: php
+   
+      $options = [
+          'limit' => 5,
+          '<option name>' => '<specification>',
+      ];
+
+This guide provides examples of how to use builders in non-aggregation
+operations. To view aggregation examples, see the :ref:`php-aggregation`
+guide.
+
+Sample Data
+~~~~~~~~~~~
+
+The examples in this guide use the ``shipwrecks`` collection in the
+``sample_geospatial`` database from the :atlas:`Atlas sample datasets
+</sample-data>`. To access this collection from your PHP application,
+instantiate a ``MongoDB\Client`` that connects to an Atlas cluster
+and assign the following value to your ``$collection`` variable:
+
+.. literalinclude:: /includes/builders.php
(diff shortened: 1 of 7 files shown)


source/builders.txt
```diff
@@ -0,0 +1,305 @@
...
+
+This example performs the following actions:
+
+- Uses the ``Stage::match()`` method to filter for only update change
+  events
+- Uses the ``Stage::project()`` method to output only the specified change
+  event fields
+- Calls the :phpmethod:`MongoDB\Collection::watch()`
+  method to open the change stream with the option to output the full
+  document after update
```
lindseymoore, line 284:
```suggestion
- Calls the :phpmethod:`MongoDB\Collection::watch()`
  method to open the change stream with the ``fullDocument`` option to output the full
  document after update
```
S: Include the option name for clarity

source/builders.txt
```diff
@@ -0,0 +1,305 @@
...
+   :keywords: aggregation, query, code example, type-safe
+
+Overview
+--------
+
+In this guide, you can learn about the **builder classes**
+that the {+library-short+} provides to create types used in your
+operations. You can use the builder classes and factory methods from the
+:ref:`php-aggregation-builder-api` feature to create filters for other
+operations such as find, update, and delete operations.
```
lindseymoore, line 27:
```suggestion
operations. You can use the builder classes and factory methods from the
aggregation builder feature to create filters for other
operations such as find, update, and delete operations. To learn more about aggregation builders,
see the :ref:`php-aggregation-builder-api` section of the {+library-short+}'s aggregation guide. 
```
S: Include link intro text
style guide: "Link text should begin with the purpose of the cross-reference"

source/builders.txt
```diff
@@ -0,0 +1,305 @@
...
+
+Using builders to create queries helps you identify errors at compile
+time and avoid them at runtime. This guide provides information on
+builder classes that you can use for the following tasks:
```
lindseymoore, line 31:
```suggestion
time and avoid them at runtime. This guide provides information on
builder classes in the following sections:
```
S: Introduce that these are links to the below sections. Consider using the same names as the section titles as well.

source/builders.txt
```diff
@@ -0,0 +1,305 @@
...
+
+- Uses the ``Query::eq()`` factory method to match documents in which
+  the ``watlev`` field value is ``'partly submerged at high water'``
+- Uses builders to specify an update that sets
+  sets the ``year`` field to ``1870``
+- Calls the :phpmethod:`MongoDB\Collection::updateOne()`
+  method to perform the update
+- Prints the number of updated documents
+
+.. io-code-block::
```
lindseymoore, line 227: S: For long code examples like this, I think code comments would be helpful to include at each section break

source/builders.txt
```diff
@@ -0,0 +1,305 @@
...
+   .. output::
+      :language: json
+      :visible: false
+
+      {
+        "_id":...,
+        "operationType":"update",
+        "fullDocument":{"_id":...,"feature_type":"Wrecks - Visible",...},
+        "ns":{"db":"sample_geospatial","coll":"shipwrecks"}
+      }
```
lindseymoore, line 305: S: And API documentation section would be helpful 
```suggestion
      }
     
Additional Information
----------------------
```

lindseymoore (changes requested):
Looks good so far! A few suggestions

source/builders.txt
```diff
@@ -0,0 +1,305 @@
...
+filter definitions to use in find, update, and delete operations. To
+learn more about creating filters, see the :ref:`php-specify-query` guide.
+
+The following steps describe how to create a filter definition by using
+builders:
+
+1. Instantiate the ``QueryInterface`` by calling the ``Query::query()``
+   method.
+   
+#. Pass the field name to compare and a factory method from the
```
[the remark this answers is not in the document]
rustagir: reworded

source/builders.txt
```diff
@@ -0,0 +1,305 @@
...
+      :visible: false
+
+      Deleted documents: 1
+
+To learn more about delete operations, see the :ref:`php-write-delete`
+guide.
+
+.. _php-builders-update:
+
+Define an Update Document
```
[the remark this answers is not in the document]
rustagir: This section isn't about performing updates, it is about using builders to create the update document

source/builders.txt
```diff
@@ -0,0 +1,305 @@
...
+
+.. important::
+    
+   The only update operation that you can express by using builders uses
+   the ``$set`` operator. This operator allows you to replace the value of
+   a field with a specified value or create a new field that has a specified
+   value. To learn more, see the :manual:`$set </reference/operator/update/set/>`
+   reference in the {+mdb-server+} manual.
+
+The following steps describe how to create an update document by using
```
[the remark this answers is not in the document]
rustagir: The builders do not actually update the document, they are just used to create the update document/specify the update. So, i believe wording should stay

source/builders.txt
```diff
@@ -0,0 +1,305 @@
...
+builders:
+
+1. Create a ``Pipeline`` instance.
+   
+#. Pass one or more ``$set`` stages by calling the ``Stage::set()``
+   method and passing a field name and a value.
+
+The following code shows the template to define an update by using
```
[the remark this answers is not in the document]
rustagir: See other comments

source/builders.txt
```diff
@@ -0,0 +1,305 @@
...
+You can use factory methods from the ``Stage`` class to modify a change
+stream's output by creating a pipeline. To learn more about change
+streams, see the :ref:`php-change-streams` guide.
+
+The following steps describe how to create a change stream filter by
+using builders:
+
+1. Create an array.
+   
+#. Pass one or more ``$match`` stages by calling factory methods from
```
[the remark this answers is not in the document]
rustagir: The PHP team is not publishing API docs for the builder methods - see the comments on https://jira.mongodb.org/browse/DOCSP-43919

The aggregation builder page has more info that says that there is a supported builder for every agg operator

source/builders.txt
rustagir: see other reply about API documentation

source/builders.txt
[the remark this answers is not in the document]
lindseymoore: I would define exactly what is meant by an update document then. I'll leave a comment as example, should be simple.

source/builders.txt
```diff
@@ -0,0 +1,306 @@
...
+.. _php-builders-update:
+
+Define an Update Document
+-------------------------
+
+You can use factory methods from the ``Stage`` builder class to update
+or create field values in your update operations. To learn more about
+updating documents, see the :ref:`php-write-update` guide.
+
+.. important:: Only $set Operator Supported
```
lindseymoore, line 192:
```suggestion
.. important:: Only $set Operator Supported for Update Operations
```
S: For clarity

source/includes/builders.php
```diff
@@ -0,0 +1,87 @@
...
+
+echo 'Updated documents: ', $result->getModifiedCount(), PHP_EOL;
+// end-updateone
+
+// start-cs
+/* Creates a pipeline to filter for update operations and returns
+   only specific fields */
+$pipeline = [
+    Stage::match(operationType: Query::eq('update')),
+    Stage::project(operationType: 1, ns: 1, fullDocument: 1),
```
lindseymoore, line 66: Q: How are users supposed to know what the options in this project stage mean? For example, what is 'ns'? Do you know which page, if any, would have this info? I am not seeing it on the Aggregation page linked to.
If no page can be linked to, I would explain what the projection does more thoroughly (but succinctly ofc)

lindseymoore (changes requested):
Few more small things

source/includes/builders.php
rustagir: The change stream guide is linked to in this section, but I can add another quick explanation

source/builders.txt
```diff
@@ -0,0 +1,310 @@
...
+
+      {
+        "_id":...,
+        "operationType":"update",
+        "fullDocument":{"_id":...,"feature_type":"Wrecks - Visible",...},
+        "ns":{"db":"sample_geospatial","coll":"shipwrecks"}
+      }
+
+To learn more about the information provided by change events, see
+:manual:`Change Events <reference/change-events/>` in the {+mdb-server+} manual.
```
lindseymoore, line 310: I: This link is 404ing

lindseymoore (approved):
1 404, otherwise LGTM, Thanks!

source/builders.txt
rustagir: was missing a slash! good catch

source/includes/builders.php
```diff
@@ -0,0 +1,87 @@
...
+        Stage::set(year: 1870),
+    ),
+);
+
+echo 'Updated documents: ', $result->getModifiedCount(), PHP_EOL;
+// end-updateone
+
+// start-cs
+/* Creates a pipeline to filter for update operations and returns
+   only specific fields */
```
GromNaN, line 63: Usually, we would use this kind of comment.
```suggestion
// Creates a pipeline to filter for update operations and returns
// only specific fields.
```

source/builders.txt
```diff
@@ -0,0 +1,310 @@
...
+---------------
+
+You can use factory methods from the ``Query`` builder class to create
+filter definitions to use in find, update, and delete operations. To
+learn more about creating filters, see the :ref:`php-specify-query` guide.
+
+The following steps describe how to create a filter definition by using
+builders:
+
+1. Instantiate the ``QueryInterface`` by calling the ``Query::query()``
```
GromNaN, line 99: I don't think we need to tell about the `QueryInterface`, that's an implementation detail.

The benefit of the `Query::query()` function is to create a query using **named argument syntax** and **type safety**.

`Query::query(foo: 'bar')` is equivalent to `['foo' => 'bar']`, but adds type contraint on the values.


For more advanced constraints using `$and` and `$or`, we don't need the "key". There is an example here: https://github.com/mongodb/mongo-php-library/blob/v1.x/tests/Builder/Query/AndOperatorTest.php

source/includes/builders.php
```diff
@@ -0,0 +1,89 @@
...
+// start-find
+// Creates a query filter by using builders and
+// retrieves matching documents
+$docs = $collection->find(Query::query(
+    feature_type: Query::eq('Wrecks - Visible'),
+    coordinates: Query::near(
+        Query::geometry(
+            type: 'Point',
+            coordinates: [-79.9, 9.3],
+        ),
```
GromNaN, line 27: Do you think a method `Geometry::point(coordinates: [-79.9, 9.3])` would be useful for this?

source/builders.txt
rustagir: I will add some explanation about the usefulness of using the Query class

rustagir: I will also change one of the examples to use and or or

source/includes/builders.php
rustagir: No, i think that users might appreciate this syntax mirroring the MQL so that they can use the server docs to get more context. So it might not be helpful to a large number of users to create this convenience method

source/builders.txt
```diff
@@ -0,0 +1,334 @@
...
+
+Define an Update Document
+-------------------------
+
+You can use the ``Stage::set()`` factory method when creating update
+documents. Update documents describe the updates to make to target
+documents. To learn more about updating documents, see the
+:ref:`php-write-update` guide.
+
+.. important:: Only $set Operator Supported in Update Documents
```
GromNaN, line 217: That's confusing. 
Update supports an [aggregation pipeline](https://www.mongodb.com/docs/manual/tutorial/update-documents-with-aggregation-pipeline/), in that case most of the stages can be used.

Or by default that's [update operators](https://www.mongodb.com/docs/manual/reference/operator/update/) that are not supported by the builder yet. They are not stages even if the syntax for `$set` is similar.

GromNaN: Yes, you can remove it. We will try to add this as a new feature: [PHPLIB-1638](https://jira.mongodb.org/browse/PHPLIB-1638)

GromNaN (approved):

PR #215, opened 2025-03-05
DOCSP-47835 Clarify batchsize behavior

shuangela:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-47835>
Staging - <https://deploy-preview-215--docs-php-library.netlify.app/reference/method/MongoDBCollection-listSearchIndexes/#parameters>
<https://deploy-preview-215--docs-php-library.netlify.app/reference/method/MongoDBDatabase-aggregate/#parameters>
<https://deploy-preview-215--docs-php-library.netlify.app/reference/method/MongoDBCollection-find/#parameters>
<https://deploy-preview-215--docs-php-library.netlify.app/reference/method/MongoDBDatabase-watch/#parameters>
<https://deploy-preview-215--docs-php-library.netlify.app/read/retrieve/#modify-find-behavior>

Note: This topic is a bit over my head in terms of technicalities of the batchsize parameter. I read the slack thread provided in the ticket as well as found the doc that I linked in the ticket about batchsize, but any suggestions about clarity or accuracy are greatly appreciated!

## Self-Review Checklist

### DIFF
--- a/source/includes/extracts-watch-option.yaml
+++ b/source/includes/extracts-watch-option.yaml
@@ -1,10 +1,15 @@
 ref: watch-option-batchSize
 content: |
-  The maximum number of documents within each batch returned in a query result, which applies 
+  The maximum number of documents within each batch returned in a change stream, which applies 
   to the ``aggregate`` command. By default, the ``aggregate`` command has an initial batch size of 
   ``101`` documents and a maximum size of 16 mebibytes (MiB) for each subsequent batch. This
   option can enforce a smaller limit than 16 MiB, but not a larger one. If you set ``batchSize`` 
   to a limit that results in batches larger than 16 MiB, this option has no effect.
+
+  Irrespective of the ``batchSize`` option, the initial ``aggregate`` command
+  response for a change stream generally does not include any documents
+  unless another option is used to configure its starting point (e.g.
+  ``startAfter``).
 ---
 ref: watch-option-fullDocument
 content: |
--- a/source/includes/extracts-watch-option.yaml
+++ b/source/includes/extracts-watch-option.yaml
@@ -2,13 +2,9 @@ ref: watch-option-batchSize
 content: |
   The maximum number of documents within each batch returned in a query result, which applies 
   to the ``aggregate`` command. By default, the ``aggregate`` command has an initial batch size of 
-  ``101`` documents and a maximum size of 16 mebibytes for each subsequent batch. This
-  option can enforce a smaller limit than 16 mebibytes, but not a larger
-  one. If you set ``batchSize`` to a limit that results in batches larger than
-  16 MiB, this option has no effect.
-
-  This command determines the maximum number of change events to return in each response from the
-  server.
+  ``101`` documents and a maximum size of 16 mebibytes (MiB) for each subsequent batch. This
+  option can enforce a smaller limit than 16 MiB, but not a larger one. If you set ``batchSize`` 
+  to a limit that results in batches larger than 16 MiB, this option has no effect.
 ---
 ref: watch-option-fullDocument
 content: |
--- a/source/includes/extracts-watch-option.yaml
+++ b/source/includes/extracts-watch-option.yaml
@@ -1,16 +1,14 @@
 ref: watch-option-batchSize
 content: |
-  The maximum number of documents within each batch returned in a query result. By default,
-  the ``find()`` method has an initial batch size of ``101`` documents
-  and a maximum of 16 megabytes for each subsequent batch. This
-  option can enforce a smaller limit than 16 megabytes, but not a larger
+  The maximum number of documents within each batch returned in a query result, which applies 
+  to the ``aggregate`` command. By default, the ``aggregate`` command has an initial batch size of 
+  ``101`` documents and a maximum size of 16 mebibytes for each subsequent batch. This
+  option can enforce a smaller limit than 16 mebibytes, but not a larger
   one. If you set ``batchSize`` to a limit that results in batches larger than
-  16 MB, this option has no effect.
+  16 MiB, this option has no effect.
 
-  Irrespective of the ``batchSize`` option, the initial ``aggregate`` command
-  response for a change stream generally does not include any documents
-  unless another option is used to configure its starting point (e.g.
-  ``startAfter``).
+  This command determines the maximum number of change events to return in each response from the
+  server.
 ---
 ref: watch-option-fullDocument
 content: |
--- a/source/includes/extracts-watch-option.yaml
+++ b/source/includes/extracts-watch-option.yaml
@@ -1,9 +1,11 @@
 ref: watch-option-batchSize
 content: |
-  Specifies the batch size for the cursor, which will apply to both the initial
-  ``aggregate`` command and any subsequent ``getMore`` commands. This determines
-  the maximum number of change events to return in each response from the
-  server.
+
+  Specifies the batch size for the cursor, which will apply to both the
+  initial ``aggregate`` command and any subsequent ``getMore`` commands.
+  If batchSize is set, ``aggregate`` and any subsequent ``getMore`` commands returns the smaller of 16 megabytes of data 
+  or batchSize documents. Operations of type ``find()``, ``aggregate()``, ``listIndexes``, and ``listCollections`` return a maximum of ``16`` megabytes per batch. 
+  batchSize can enforce a smaller limit than ``16`` megabytes, but not a larger one. 
 
   Irrespective of the ``batchSize`` option, the initial ``aggregate`` command
   response for a change stream generally does not include any documents
--- a/source/includes/extracts-watch-option.yaml
+++ b/source/includes/extracts-watch-option.yaml
@@ -4,7 +4,7 @@ content: |
   Specifies the batch size for the cursor, which will apply to both the
   initial ``aggregate`` command and any subsequent ``getMore`` commands.
   If batchSize is set, ``aggregate`` and any subsequent ``getMore`` commands returns the smaller of 16 megabytes of data 
-  or batchSize documents. Operations of type ``find()``, ``aggregate()``, ``listIndexes``, and ``listCollections`` return a maximum of ``16`` megabytes per batch. 
+  or batchSize documents. Operations of type ``find``, ``aggregate``, ``listIndexes``, and ``listCollections`` return a maximum of ``16`` megabytes per batch. 
   batchSize can enforce a smaller limit than ``16`` megabytes, but not a larger one. 
 
   Irrespective of the ``batchSize`` option, the initial ``aggregate`` command
--- a/source/includes/extracts-watch-option.yaml
+++ b/source/includes/extracts-watch-option.yaml
@@ -1,11 +1,11 @@
 ref: watch-option-batchSize
 content: |
-
-  Specifies the batch size for the cursor, which will apply to both the
-  initial ``aggregate`` command and any subsequent ``getMore`` commands.
-  If batchSize is set, ``aggregate`` and any subsequent ``getMore`` commands returns the smaller of 16 megabytes of data 
-  or batchSize documents. Operations of type ``find``, ``aggregate``, ``listIndexes``, and ``listCollections`` return a maximum of ``16`` megabytes per batch. 
-  batchSize can enforce a smaller limit than ``16`` megabytes, but not a larger one. 
+  The maximum number of documents within each batch returned in a query result. By default,
+  the ``find()`` method has an initial batch size of ``101`` documents
+  and a maximum of 16 megabytes for each subsequent batch. This
+  option can enforce a smaller limit than 16 megabytes, but not a larger
+  one. If you set ``batchSize`` to a limit that results in batches larger than
+  16 MB, this option has no effect.
 
   Irrespective of the ``batchSize`` option, the initial ``aggregate`` command
   response for a change stream generally does not include any documents
--- a/source/read/retrieve.txt
+++ b/source/read/retrieve.txt
@@ -182,10 +182,10 @@ you can set in the array:
    * - ``batchSize`` 
      - | The maximum number of documents within each batch returned in a query result. By default,
          the ``find()`` method has an initial batch size of ``101`` documents
-         and a maximum of 16 megabytes for each subsequent batch. This
-         option can enforce a smaller limit than 16 megabytes, but not a larger
+         and a maximum size of 16 mebibytes for each subsequent batch. This
+         option can enforce a smaller limit than 16 mebibytes, but not a larger
          one. If you set ``batchSize`` to a limit that results in batches larger than
-         16 MB, this option has no effect.
+         16 MiB, this option has no effect.
        | **Type**: ``integer``
 
    * - ``collation`` 
--- a/source/read/retrieve.txt
+++ b/source/read/retrieve.txt
@@ -181,7 +181,7 @@ you can set in the array:
 
    * - ``batchSize`` 
      - | The maximum number of documents within each batch returned in a query result. By default,
-         the ``find()`` method has an initial batch size of ``101`` documents
+         the ``find`` command has an initial batch size of ``101`` documents
          and a maximum size of 16 mebibytes (MiB) for each subsequent batch. This
          option can enforce a smaller limit than 16 MiB, but not a larger
          one. If you set ``batchSize`` to a limit that results in batches larger than
--- a/source/read/retrieve.txt
+++ b/source/read/retrieve.txt
@@ -182,8 +182,8 @@ you can set in the array:
    * - ``batchSize`` 
      - | The maximum number of documents within each batch returned in a query result. By default,
          the ``find()`` method has an initial batch size of ``101`` documents
-         and a maximum size of 16 mebibytes for each subsequent batch. This
-         option can enforce a smaller limit than 16 mebibytes, but not a larger
+         and a maximum size of 16 mebibytes (MiB) for each subsequent batch. This
+         option can enforce a smaller limit than 16 MiB, but not a larger
          one. If you set ``batchSize`` to a limit that results in batches larger than
          16 MiB, this option has no effect.
        | **Type**: ``integer``
--- a/source/read/retrieve.txt
+++ b/source/read/retrieve.txt
@@ -180,7 +180,9 @@ you can set in the array:
      - Description
 
    * - ``batchSize`` 
-     - | The number of documents to return per batch. The default value is ``101``.
+     - | The limit for each batch returned in a query result. The ``find()`` method has an initial batch size of 101 documents by default and returns a maximum of 16
+         megabytes per batch. Subsequent ``getMore`` operations issued against the resulting cursor have no default batch size, 
+         so they are limited only by the ``16`` megabyte message size. batchSize can enforce a smaller limit than ``16`` megabytes, but not a larger one. 
        | **Type**: ``integer``
 
    * - ``collation`` 
--- a/source/read/retrieve.txt
+++ b/source/read/retrieve.txt
@@ -180,9 +180,12 @@ you can set in the array:
      - Description
 
    * - ``batchSize`` 
-     - | The limit for each batch returned in a query result. The ``find`` method has an initial batch size of ``101`` documents by default and returns a maximum of 16
-         megabytes per batch. Subsequent ``getMore`` operations issued against the resulting cursor have no default batch size, 
-         so they are limited only by the ``16`` megabyte message size. batchSize can enforce a smaller limit than ``16`` megabytes, but not a larger one. 
+     - | The maximum number of documents within each batch returned in a query result. By default,
+         the ``find()`` method has an initial batch size of ``101`` documents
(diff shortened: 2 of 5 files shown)


rustagir (changes requested):
I have left some comments that should give context to revise this PR

source/includes/extracts-watch-option.yaml
```diff
@@ -1,9 +1,11 @@
...
 content: |
-  Specifies the batch size for the cursor, which will apply to both the initial
-  ``aggregate`` command and any subsequent ``getMore`` commands. This determines
-  the maximum number of change events to return in each response from the
-  server.
+
+  Specifies the batch size for the cursor, which will apply to both the
+  initial ``aggregate`` command and any subsequent ``getMore`` commands.
+  If batchSize is set, ``aggregate`` and any subsequent ``getMore`` commands returns the smaller of 16 megabytes of data 
+  or batchSize documents. Operations of type ``find``, ``aggregate``, ``listIndexes``, and ``listCollections`` return a maximum of ``16`` megabytes per batch.
```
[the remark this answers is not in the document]
shuangela: Question: to my knowledge, the default size that applies to the initial command (101 documents) and the default size (up to 16 mb) for requests for more batches is different. Should I clarify that as well? I was a little confused on the sections about batchsize in the find() and aggregate() function pages especially, because those to my knowledge are only initial commands (as getmore is the request for more batches). So in the find() and aggregate() pages, do you think it makes sense to only talk about the initial command and not the requests for more batches?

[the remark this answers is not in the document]
rustagir: I think it works like this - 
- default setting - 101 documents for first batch, 16 mb limit for subsequent
- custom setting (lets say 60) - 60 documents in first batch, 60 documents in subsequent batches as long as the total data size does not exceed 16 mb

When using the driver, users **do not** need to manually run a `getMore` command. The driver takes care of this internally. Thus, even though it is under the hood, the batch size would still apply to both the first and subsequent batches AFAIK. @jmikola do you have any insight into this clarification?

source/includes/extracts-watch-option.yaml
```diff
@@ -1,9 +1,11 @@
...
-  Specifies the batch size for the cursor, which will apply to both the initial
-  ``aggregate`` command and any subsequent ``getMore`` commands. This determines
-  the maximum number of change events to return in each response from the
-  server.
+  The maximum number of documents within each batch returned in a query result. By default,
+  the ``find()`` method has an initial batch size of ``101`` documents
+  and a maximum of 16 megabytes for each subsequent batch. This
+  option can enforce a smaller limit than 16 megabytes, but not a larger
+  one. If you set ``batchSize`` to a limit that results in batches larger than
+  16 MB, this option has no effect.
```
jmikola, line 8: Change streams have nothing to do with the `find()` method. These use the [`aggregate` command](https://www.mongodb.com/docs/manual/reference/command/aggregate/) and `$changeStream` stage. A batch size for `aggregate` is expressed via the `cursor.batchSize` option on the outgoing command. For [`getMore` commands](https://www.mongodb.com/docs/manual/reference/command/getMore/) (common to both `aggregate` and `find`), `batchSize` _is_ a top-level option.

I don't think we need to repeat the server's default value within driver docs. It should be sufficient to say that the server's default will apply.

In the interest of correctness and consistency (with other docs), we should refer to "mebibytes" or "MiB" instead of "megabytes" and "MB".

source/reference/method/MongoDBCollection-find.txt
```diff
@@ -52,9 +52,12 @@ Parameters
 
      * - batchSize
        - integer
-       - The number of documents to return in the first batch. Defaults to
-         ``101``. A batchSize of ``0`` means that the cursor will be
-         established, but no documents will be returned in the first batch.
```
jmikola, line 57: The special behavior of `batchSize: 0` is notable, and also discussed in the [`find` command](https://www.mongodb.com/docs/manual/reference/command/find/) docs. Particularly with `aggregate` operations, there is a legitimate use case for wanting to start the aggregation on the server without waiting for the first result to become available. Allowing a cursor to be created immediately and iterated later (via `batchSize: 0` satisfies that).

source/reference/method/MongoDBDatabase-aggregate.txt
```diff
@@ -54,11 +54,13 @@ Parameters
 
      * - batchSize
        - integer
-       - Specifies the batch size for the cursor, which will apply to both the
-         initial ``aggregate`` command and any subsequent ``getMore`` commands.
-         This determines the maximum number of documents to return in each
-         response from the server.
```
jmikola, line 60: I think we can discuss `batchSize: 0` here as we do for the `find()` option, since there is a legitimate use case for doing so.

There is a special directive in the [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.md) for drivers to intentionally omit `batchSize` for pipelines that include `$out` or `$merge`, since it might prevent the pipeline from executing (and writing output). I don't think we need to mention that, though, as it is enforced internally (see: [Aggregate.php](https://github.com/mongodb/mongo-php-library/blob/1.21.1/src/Operation/Aggregate.php#L204-L209)).

source/includes/extracts-watch-option.yaml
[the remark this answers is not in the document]
jmikola: > Also, by total data size, do you mean added up the total size of all of the batches, or that the size of each individual batch has to be under 16 mb but the total size of batches combined can be > 16 mb? I interpreted it as each individual batch has to be under 16 mb but the total size doesn't matter.

AFAIK, there is no documented restriction on the total size of _all_ batches. The 16 MiB limit applies to each batch, and is discussed in the `batchSize` docs for the [`getMore` command](https://www.mongodb.com/docs/manual/reference/command/getMore/).

I can confirm that applications do not manually run `getMore` commands. Drivers take care of that when iterating a cursor.

@rustagir's understanding about the default and custom setting in [this comment](https://github.com/mongodb/docs-php-library/pull/215#discussion_r1983861613) is correct.

Relevant docs within drivers for this would `batchSize` within the AggregateOptions and FindOptions sections of our [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.md), which talk about inheriting the option for subsequent `getMore` commands.

Additionally, our [find/getMore/killCursors spec](https://github.com/mongodb/specifications/blob/master/source/find_getmore_killcursors_commands/find_getmore_killcursors_commands.md#behavior-of-limit-skip-and-batchsize) goes into more detail about how `batchSize` interacts with `limit` and `skip`. For example, when a `limit` is specified drivers may use a calculated value for `batchSize` to ensure we don't return more documents than necessary _and_ guarantee that the server closes the cursor. This isn't something we need to rehash in driver docs, though, since it's handled internally. Furthermore, the server docs don't seem to talk about that either (in [`cursor.limit()`](https://www.mongodb.com/docs/manual/reference/method/cursor.limit/) and [`cursor.batchSize()`](https://www.mongodb.com/docs/manual/reference/method/cursor.batchSize/)).

source/includes/extracts-watch-option.yaml
jmikola: Adding on to this, I think it'd be more technically correct to refer to "`aggregate` command" instead of "`watch()` method". That helps communicate to users that this driver method is actually running an `aggregate` command on the server, and I think it's helpful since they can then cross-reference with the server's own API docs for commands.

jmikola:
For additional context, I want to cite the docs from [Cursor Batches](https://www.mongodb.com/docs/manual/tutorial/iterate-a-cursor/#cursor-batches), which the OP in [DOCSP-47826](https://jira.mongodb.org/browse/DOCSP-47826) suggested was most clear on the subject:

> Operations of type [find()](https://www.mongodb.com/docs/manual/reference/method/db.collection.find/#mongodb-method-db.collection.find), [aggregate()](https://www.mongodb.com/docs/manual/reference/method/db.collection.aggregate/#mongodb-method-db.collection.aggregate), [listIndexes](https://www.mongodb.com/docs/manual/reference/command/listIndexes/#mongodb-dbcommand-dbcmd.listIndexes), and [listCollections](https://www.mongodb.com/docs/manual/reference/command/listCollections/#mongodb-dbcommand-dbcmd.listCollections) return a maximum of 16 megabytes per batch. [batchSize()](https://www.mongodb.com/docs/manual/reference/method/cursor.batchSize/#mongodb-method-cursor.batchSize) can enforce a smaller limit, but not a larger one.
>
> find() and aggregate() operations have an initial batch size of 101 documents by default. Subsequent [getMore](https://www.mongodb.com/docs/manual/reference/command/getMore/#mongodb-dbcommand-dbcmd.getMore) operations issued against the resulting cursor have no default batch size, so they are limited only by the 16 mebibyte message size.

source/includes/extracts-watch-option.yaml
shuangela: @jmikola For clarification, in this comment "Adding on to this, I think it'd be more technically correct to refer to "`aggregate` command" instead of "`watch()` method". That helps communicate to users that this driver method is actually running an aggregate command on the server, and I think it's helpful since they can then cross-reference with the server's own API docs for commands," when you say the "`watch()` method" did you mean the "`find()` method"? I don't think this docs refer to the watch() method. Just wanted to clarify to be sure I'm understanding correctly!

jmikola: This file is `extracts-watch-option.yaml` and included in the docs for `watch()` methods on Client, Database, and Collection objects.

source/includes/extracts-watch-option.yaml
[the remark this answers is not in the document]
jmikola: I expect the references to "MB" in the existing PHP docs are just old, and could pre-date when the server manual started using "MiB" consistently. There is no reason these can't all be changed to "MiB", although that needn't be done in this PR.

Anything relevant to this PR should use "MiB", though, so we stop introducing more instances of "MB".

source/includes/extracts-watch-option.yaml
shuangela: Ahh, I think I misunderstood the comment and understand what you're saying now. Thanks for the help!

source/includes/extracts-watch-option.yaml
```diff
@@ -1,14 +1,14 @@
 ref: watch-option-batchSize
 content: |
-  Specifies the batch size for the cursor, which will apply to both the initial
-  ``aggregate`` command and any subsequent ``getMore`` commands. This determines
-  the maximum number of change events to return in each response from the
+  The maximum number of documents within each batch returned in a query result, which applies 
+  to the ``aggregate`` command. By default, the ``aggregate`` command has an initial batch size of 
+  ``101`` documents and a maximum size of 16 mebibytes for each subsequent batch. This
```
rustagir, line 5:
```suggestion
  ``101`` documents and a maximum size of 16 mebibytes (MiB) for each subsequent batch. This
```

source/includes/extracts-watch-option.yaml
```diff
@@ -1,14 +1,14 @@
...
+  option can enforce a smaller limit than 16 mebibytes, but not a larger
```
rustagir, line 6:
```suggestion
  option can enforce a smaller limit than 16 MiB, but not a larger
```

source/includes/extracts-watch-option.yaml
```diff
@@ -1,14 +1,14 @@
...
+  one. If you set ``batchSize`` to a limit that results in batches larger than
+  16 MiB, this option has no effect.
+
+  This command determines the maximum number of change events to return in each response from the
   server.
```
rustagir, line 11: Q: should you also remove this sentence?

source/read/retrieve.txt
```diff
@@ -180,7 +180,12 @@ you can set in the array:
      - Description
 
    * - ``batchSize`` 
-     - | The number of documents to return per batch. The default value is ``101``.
+     - | The maximum number of documents within each batch returned in a query result. By default,
+         the ``find()`` method has an initial batch size of ``101`` documents
+         and a maximum size of 16 mebibytes for each subsequent batch. This
+         option can enforce a smaller limit than 16 mebibytes, but not a larger
+         one. If you set ``batchSize`` to a limit that results in batches larger than
+         16 MiB, this option has no effect.
```
rustagir, line 188:
```suggestion
         and a maximum size of 16 mebibytes (MiB) for each subsequent batch. This
         option can enforce a smaller limit than 16 MiB, but not a larger
         one. If you set ``batchSize`` to a limit that results in batches larger than
         16 MiB, this option has no effect.
```

source/reference/method/MongoDBCollection-find.txt
```diff
@@ -52,9 +52,15 @@ Parameters
...
+       - The maximum number of documents to return in the first batch. By default, the ``find()`` 
+         command has an initial batch size of ``101`` documents. This
+         option can enforce a smaller limit than 16 mebibytes, but not a larger
+         one. If you set ``batchSize`` to a limit that results in batches larger than
+         16 MiB, this option has no effect.
```
rustagir, line 59:
```suggestion
         command has an initial batch size of ``101`` documents. This
         option can enforce a smaller limit than 16 mebibytes (MiB), but not a larger
         one. If you set ``batchSize`` to a limit that results in batches larger than
         16 MiB, this option has no effect.
```

source/reference/method/MongoDBCollection-find.txt
```diff
@@ -52,9 +52,15 @@ Parameters
...
+
+         A batchSize of ``0`` means that the cursor will be established, but no documents 
+         will be returned in the first batch. This may be useful for quickly returning a cursor
+         or failure from ``aggregate`` without doing significant server-side work.
```
rustagir, line 63: I: remove ``aggregate`` from this description as its about the find method. I think JM meant that setting this option is useful for aggregation contexts but that users might use it for finds too

rustagir (changes requested):
a few small fixes but I think the changes look good!

source/includes/extracts-watch-option.yaml
```diff
@@ -1,14 +1,14 @@
 ref: watch-option-batchSize
 content: |
-  Specifies the batch size for the cursor, which will apply to both the initial
-  ``aggregate`` command and any subsequent ``getMore`` commands. This determines
-  the maximum number of change events to return in each response from the
+  The maximum number of documents within each batch returned in a query result, which applies
```
jmikola, line 3: Consider replacing "query result" with "change stream" to make this more contextual:

> The maximum number of documents within each batch returned in a change stream

This pairs with my suggestion to remove a sentence below. Alternatively, you can try to work "change events" into this sentence, but I think having both "change events" and "change stream" is too repetitive.

source/reference/method/MongoDBCollection-find.txt
```diff
@@ -52,9 +52,15 @@ Parameters
 
      * - batchSize
        - integer
-       - The number of documents to return in the first batch. Defaults to
-         ``101``. A batchSize of ``0`` means that the cursor will be
-         established, but no documents will be returned in the first batch.
+       - The maximum number of documents to return in the first batch. By default, the ``find()``
```
jmikola, line 55: Not sure why this leading sentence differs so much from the others. Also, the command should be written as `find`. `find()` would refer to the helper method in the driver.

source/includes/extracts-watch-option.yaml
shuangela: Good call out, I'll remove!

source/reference/method/MongoDBCollection-find.txt
```diff
@@ -52,9 +52,15 @@ Parameters
 
      * - batchSize
        - integer
-       - The number of documents to return in the first batch. Defaults to
-         ``101``. A batchSize of ``0`` means that the cursor will be
-         established, but no documents will be returned in the first batch.
+       - The maximum number of documents to return in the first batch. By default, the ``find()`` 
+         command has an initial batch size of ``101`` documents. This
```
jmikola, line 56: This should definitely specify:

> ...and a maximum size of 16 mebibytes for each subsequent batch

As-is, it's inconsistent with the other docs snippets and could mislead users into thinking this has no effect on `getMore` for subsequent batches.

source/reference/method/MongoDBCollection-listSearchIndexes.txt
```diff
@@ -40,16 +40,16 @@ Parameters
...
+       - The maximum number of documents within each batch returned in a query result, which applies 
+         to the ``aggregate`` command. By default, the ``aggregate`` command has an initial batch size of 
+         ``101`` documents and a maximum size of 16 mebibytes for each subsequent batch. This
+         option can enforce a smaller limit than 16 mebibytes, but not a larger
+         one. If you set ``batchSize`` to a limit that results in batches larger than
+         16 MiB, this option has no effect.
+
+         A batchSize of ``0`` means that the cursor will be established, but no documents 
+         will be returned in the first batch. This may be useful for quickly returning a cursor
+         or failure from ``aggregate`` without doing significant server-side work.
```
jmikola, line 52: I don't think we need to discuss "batchSize of 0" for the `listSearchIndexes()` helper. Consider removing this paragraph.

It makes more sense for general `aggregate()` helpers and to a lesser extent `find()`. But the fact that listing search indexes uses `aggregate` is really more of an implementation detail that users won't care about.

source/includes/extracts-watch-option.yaml
```diff
@@ -1,14 +1,14 @@
...
+  one. If you set ``batchSize`` to a limit that results in batches larger than
+  16 MiB, this option has no effect.
+
+  This command determines the maximum number of change events to return in each response from the
   server.
-
-  Irrespective of the ``batchSize`` option, the initial ``aggregate`` command
-  response for a change stream generally does not include any documents
-  unless another option is used to configure its starting point (e.g.
-  ``startAfter``).
```
jmikola, line 11: I didn't notice this earlier, but this paragraph seems like a relevant bit of information to preserve.

source/includes/extracts-watch-option.yaml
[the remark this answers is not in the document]
shuangela: Yup, Rea also called this out and will implement the fix!

source/includes/extracts-watch-option.yaml
```diff
@@ -1,14 +1,14 @@
 ref: watch-option-batchSize
 content: |
-  Specifies the batch size for the cursor, which will apply to both the initial
-  ``aggregate`` command and any subsequent ``getMore`` commands. This determines
-  the maximum number of change events to return in each response from the
+  The maximum number of documents within each batch returned in a query result, which applies 
+  to the ``aggregate`` command. By default, the ``aggregate`` command has an initial batch size of
```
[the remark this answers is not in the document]
shuangela: I did remove mention of GetMore because of the discussion earlier in the PR about how GetMore in the PHP driver is under the hood/not manually ran by the user. I see what you're saying about mentioning each subsequent batch in all docs, I'll go back and make sure that those mentions exist.

source/reference/method/MongoDBCollection-listSearchIndexes.txt
```diff
@@ -40,16 +40,12 @@ Parameters
...
-         initial ``aggregate`` command; subsequent ``getMore`` commands will use
-         the server's default batch size. This may be useful for quickly
-         returning a cursor or failure from ``aggregate`` without doing
-         significant server-side work.
+       - The maximum number of documents within each batch returned in a query result, which applies 
+         to the ``aggregate`` command. By default, the ``aggregate`` command has an initial batch size of 
+         ``101`` documents and a maximum size of 16 mebibytes (MiB) for each subsequent batch. This
+         option can enforce a smaller limit than 16 MiB, but not a larger
+         one. If you set ``batchSize`` to a limit that results in batches larger than
+         16 MiB, this option has no effect.
```
rustagir, line 48:
```suggestion
       - The maximum number of documents within each batch returned in the indexes list, which applies 
         to the ``aggregate`` command. By default, the ``aggregate`` command has an initial batch size of 
         ``101`` documents and a maximum size of 16 mebibytes (MiB) for each subsequent batch. This
         option can enforce a smaller limit than 16 MiB, but not a larger
         one. If you set ``batchSize`` to a limit that results in batches larger than
         16 MiB, this option has no effect.
```

rustagir (approved):
lgtm! good work on this technical update

source/read/retrieve.txt
```diff
@@ -180,7 +180,12 @@ you can set in the array:
      - Description
 
    * - ``batchSize`` 
-     - | The number of documents to return per batch. The default value is ``101``.
+     - | The maximum number of documents within each batch returned in a query result. By default,
+         the ``find()`` method has an initial batch size of ``101`` documents
```
jmikola, line 184: Since this section is talking about both `find()` and `findOne()`, I think this can just refer to "the `find` command" (used by both helpers).

Therefore, this snippet will just duplicate the first paragraph from the `batchSize` docs in `MongoDBCollection-find.txt`.

jmikola (approved):
Defer to you if "`find()` method" is better changed to "`find` command".

I'll leave a preemptive LGTM.

PR #148, opened 2024-09-23
DOCSP-41973 Read Landing Page

lindseymoore:
# Pull Request Info

[PR Reviewing Guidelines](https://github.com/mongodb/docs-php-library/blob/master/REVIEWING.md)

JIRA - <https://jira.mongodb.org/browse/DOCSP-41973>
Staging - <https://deploy-preview-148--docs-php-library.netlify.app/read/>

## Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST?
- [x] Did you run a spell-check?
- [x] Did you run a grammar-check?
- [x] Are all the links working?
- [x] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?

PR #264, opened 2025-06-04
DOCSP-43912: Improve change stream code examples

mcmorisi:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-43912>

### Staging Links

* https://deploy-preview-264--docs-php-library.netlify.app/read/
* https://deploy-preview-264--docs-php-library.netlify.app/read/change-streams/

## Self-Review Checklist

- [ ] Did you run a spell-check?
- [ ] Did you run a grammar-check?


lindseymoore (approved):
Format LGTM!

GromNaN (changes requested):
The `change-streams.php` cannot be run on its own as each change stream loop is an infinite loop (unless you drop the collection). Wouldn't is better to create separate files that we can run?

source/includes/read/change-streams/change-stream.php
```diff
@@ -0,0 +1,39 @@
...
+} while (! $changeStream->valid() || $changeStream->current()['operationType'] !== 'invalidate');
+// end-open-change-stream
+
+// Updates a document that has a "name" value of "Blarney Castle"
+// start-update-for-change-stream
+$result = $collection->updateOne(
+    ['name' => 'Blarney Castle'],
+    ['$set' => ['cuisine' => 'Irish']]
+);
+// end-update-for-change-stream
```
GromNaN, line 39: Missing newline at EOF.

```suggestion
// end-update-for-change-stream

```

source/includes/read/change-streams/change-stream-options.php
```diff
@@ -0,0 +1,32 @@
...
+$changeStream->rewind();
+
+do {
+    $changeStream->next();
+
+    if ($changeStream->valid()) {
+        $event = $changeStream->current();
+        echo toJSON($event), PHP_EOL;
+    }
+} while (! $changeStream->valid() || $changeStream->current()['operationType'] !== 'invalidate');
```
GromNaN, line 31: On reflection, this notation is much more complex than the previous example.

In my [original comment](https://github.com/mongodb/docs-php-library/pull/148#discussion_r1777214881) the idea was rather to use `while(true)` like this:

```suggestion
while (true) {
    $changeStream->next();

    if (! $changeStream->valid()) {
        continue;
    }

    $event = $changeStream->current();
    echo toJSON($event), PHP_EOL;

    if ($changeStream->current()['operationType'] === 'invalidate') {
        break;
    }
}
```

GromNaN (commented):
I'm not convinced by the new style. Keeping the previous `for` loop, or replacing with a `while(true)` would be more comprehensible.

GromNaN (approved):

